PR #2, opened 2018-08-23
WIP: Use ci-helpers to simplify the conda set-up

astrofrog:
Also switch to codecov for coverage since it produces nicer results

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,44 +7,38 @@ env:
   - PYTHON_VERSION=2.7
   - PYTHON_VERSION=3.5
   - PYTHON_VERSION=3.6
+ global:
+  - CONDA_VERSION=4.5.10
+  - PIP_DEPENDENCIES="codecov"
+  - SETUP_XVFB=True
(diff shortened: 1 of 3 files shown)


astrofrog:
Wow conda has got really slow in the last few months...

mbargull:
That `ci-helpers` thingy sets `channel_priority` to `false` which is usually not what you want to do for multiple big channels like `defaults` + `conda-forge`. I fear, you're giving the solver an extra hard time there...
If you can narrow the setup where you observe that bad performance down to a simple `conda create -c channel1 -c channel2 ... packages` which I can run from a clean Miniconda installation, we could track this on the Conda issue tracker.

astrofrog:
@mbargull - thanks for the tip on channel priority! That does seem to help.

astrofrog:
This doesn't un-break the build but it simplifies the Travis config a little. Currently there are real test failures on MacOS X, and there are issues installing ipyvolume on Linux.

PR #23, opened 2018-08-28
Update environment.yml

eteq:
This now works locally.

Two modifications that *might* not be desirable:
* I changed the `==` dependencies to `>=`.  Seems safer for the future
* Added "glueviz" as the default conda channel

cc @maartenbreddels @astrofrog


maartenbreddels:
Building on travis is a bit flaky atm, it seems ipyvolume is using quite some memory when building, so I consider it .

PR #39, opened 2018-09-26
WIP: Vaex/BaseData support for bqplot histogram and scatter2d

maartenbreddels:
With this PR we should be able to use vaex, using this package: https://github.com/glue-viz/glue-vaex


maartenbreddels:
Merging anyway, since the failure is unrelated: https://github.com/glue-viz/glue/issues/1898

Issue #9, opened 2018-08-26
Implement support for square pixel aspect ratio in image viewer


astrofrog:
Currently the image viewer doesn't constrain the aspect ratio of pixels:

<img width="749" alt="screen shot 2018-08-26 at 12 53 34" src="https://user-images.githubusercontent.com/314716/44630654-25f26600-a92f-11e8-9160-d0e2e07349d7.png">

It would be nice to support this to avoid 'squashed' images.

maartenbreddels:
In principle, an aspect of 1 would work, even without the referenced fix. Do you see a use case for aspect != 1?

astrofrog:
@maartenbreddels - I don't see any use case for a fixed aspect != 1, but we do want to allow the 'auto' aspect mode meaning the image just fills the axes (the current behavior), as this is useful when the image isn't a 'spatial' image (e.g. a PV diagram).

maartenbreddels:
yeah, so a checkbox/toggle button with 'equal aspect' would do right?

astrofrog:
@maartenbreddels - yep!

maartenbreddels:
Solved by #40

PR #40, opened 2018-10-10
Add support for equal aspect for the bqplot image viewer, solves #9

maartenbreddels:
Demo:
demo-equal-aspect

### DIFF
--- a/glue_jupyter/bqplot/view.py
+++ b/glue_jupyter/bqplot/view.py
@@ -238,6 +238,20 @@ class BqplotImageView(BqplotBaseView):
     _state_cls = ImageViewerState
     large_data_size = 2e7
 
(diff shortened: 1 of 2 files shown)


glue_jupyter/bqplot/view.py
```diff
@@ -256,6 +270,16 @@ def get_subset_layer_artist(self, layer=None, layer_state=None):
         self._add_layer_tab(layer)
         return layer
 
+    def create_tab(self):
+        super(BqplotImageView, self).create_tab()
+        self.widgets_aspect = widgets.Checkbox(description='Equals aspect')
```
astrofrog: I think 'Equal aspect ratio' would be clearer

PR #41, opened 2018-10-10
Implement show axes for bqplot

maartenbreddels:
A feature request in glue (https://github.com/glue-viz/glue/issues/1591) and was already enabled for the ipyvolume viewer, and now enabled for the bqplot viewers as well.
It would be nice to put in the `show_axes` somewhere in the state.

This seems to exposes a bug for the webgl accelerated parts in bqplot (image viewer and scattermega), it seems to work fine for the histogram. So we could merge this, or keep open until that is fixed.

demo-show-axes


astrofrog:
Nice - I agree we should add that to the state in glue.

maartenbreddels:
Ignoring py27 again

PR #42, opened 2018-10-10
Include dataset label in tab labels, resolves #11

### LINKED ISSUE #11: Subset tab name should include dataset name
In the following example, we should include the dataset name in brackets after ``Subset 1``:

screen shot 2018-08-26 at 13 07 09

maartenbreddels:
I've used the <dataset>:<subset> format, instead of putting dataset behind the subset, which to me feel more natural.
screen shot 2018-10-10 at 11 11 14


maartenbreddels:
Failure is python2.7 due to Jupyter Lab.

PR #43, opened 2018-10-10
Subset controls and selection mode (replace/and/or) more like glue-qt with glue icons

maartenbreddels:
I've made the active subset control a but like the glue-qt UI, but now using a pulldown menu where you can select multiple subsets. I'm not sure if this UI, or the select box is a better UI, I've left both in for the moment (see demo).

demo-active-subset


astrofrog:
Looks nice! I quite like the drop-down menu, and was wondering whether when you create the app object the layout could be:

```
Active subset: [...]     Selection mode: [...]
```

so that it all just takes one line?

maartenbreddels:
Yes, that make sense, more like a toolbar:
screen shot 2018-10-11 at 09 53 32

Also, now the application object is now shown by default.

maartenbreddels:
@astrofrog I see two issues you probably know more about: glue cannot seem to find its own icons, have you ever seen this? And the Error message for the Python2.7 version seems to show it is actually using Python3.7.

astrofrog:
I think the Python 2.7 issue seems to be because conda decides to update python when updating from an environment file. I'll investigate how to avoid that.

maartenbreddels:
@astrofrog I plan to do a slightly different approach with testing, in this PR I have added for instance

`glue_jupyter/widgets/subset_mode.py`

and the unittest in

`glue_jupyter/widgets/subset_mode_test.py`

It is pretty non-standard, but I find it very convenient for developing, and exploring a package. The code and tests are near each other (alphabetically, therefore ending in _test.py) and tests are therefore much easier to find. Any objection to this?

maartenbreddels:
Now with https://github.com/maartenbreddels/ipymaterialui it looks like this:
demo-glue-jupyter-mui

maartenbreddels:
Merging for now, we can always move the unittests later.

Issue #36, opened 2018-09-18
**state should be possible when creating views


eteq:
Continuing in a series of issues that are "confusing elements of the API" from work on #28 :

It's useful programatically to be able to set the state all at once on view-creation.  The natural way it seems to me to implement this is that the view-creation functions (e.g. `app.scatter2d`) should accept kwargs that then get applied directly as state.  Alternatively, a state dictionary could be passed in, but that feels a bit less natural/easy to use.

Issue #37, opened 2018-09-18
scatter*d should have color/scale options


eteq:
Continuing in a series of issues that are "confusing elements of the API" from work on #28 :

The `scatter2d` and `scatter3d` functions should have color, size, and marker options.  This is related to #33 and #34 but those are about *after the fact*, whereas this issue is about on-creation.  Also potentially connected to #36 since that could be the underlying mechanism... but the key point of this issue is that it should be possible.

PR #44, opened 2018-10-11
Initial values for plotting method

maartenbreddels:
cc @eteq 
This is the start to address https://github.com/glue-viz/glue-jupyter/issues/36 and to become a more user friendly API.
The most common arguments should be included as method arguments, other state values can be passed through the viewer_state dict or layer_state dict.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -6,7 +6,7 @@ dependencies:
   - nomkl
   - nodejs
   - pip:
-    - jupyterlab >=0.34
+    # - jupyterlab >=0.34  # since we still support python2.7, do not include this
     - notebook >=5.5
     - ipywidgets >=7.4.0
     - ipympl
--- a/glue_jupyter/__init__.py
+++ b/glue_jupyter/__init__.py
@@ -12,6 +12,8 @@ from IPython.display import display
 from glue.core.link_helpers import LinkSame
 import ipywidgets as widgets
 import six
+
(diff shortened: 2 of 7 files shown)


maartenbreddels:
Also related: https://github.com/glue-viz/glue-jupyter/issues/37

Now you can do:
```python
s = app.scatter2d('x', 'y', color='red')
```
To change it after the fact:
```python
s.layers[0].state.color = 'green'
```
The first is an improvement, the second was already possible, but I don't see a better way of doing this, unless we return the layer as well.
```python
viewer, layer = app.scatter2d('x', 'y', color='red')
layer.state.color = 'green'
```

In case you find (using tab completion of course, not reading documentation), that the state object also has an `alpha` property, which is not an argument (yet) of scatter2d, you can do:
```python
s = app.scatter2d('x', 'y', color='red', layer_state={'alpha': 0.2})
```
I think alpha should become an argument of scatter2d, but this is just to give a clear example.

If this seems like a good approach, let me know, and I'll continue this.

maartenbreddels:
And now we can also do:
```python
s = app.scatter2d('x', 'y', color='red', layer_state={'alpha': 0.2})
layer2 = s.add_data(data2, color='red', size=1, alpha=0.4)
```

maartenbreddels:
Merging this for now, is adds the basic functionality, and we can always tweak/change later. Also, ignoring py2 failures.

PR #45, opened 2018-10-19
Implement cmap_mode=Linear for bqplot scatter plot

maartenbreddels:
depends on an upcoming fix in the scatter_webgl branch of bqplot


maartenbreddels:
Ignoring the py27 failures for now

PR #50, opened 2018-11-21
Move to using ipyastroimage

maartenbreddels:
Use https://github.com/glue-viz/ipyastroimage for the bqplot Image mark, instead of trying to get all we want in bqplot.

Since we want astronomy related parts in the image mark, and want to move fast, it makes more sense to continue development outside of bqplot.


maartenbreddels:
py27 issue should be fixed in #51 51

mwcraig:
Thanks, this looks very promising!

PR #64, opened 2019-02-21
Add main tutorial notebooks and provide a require_data function

astrofrog:
This tidies up the notebooks directory, adding the main tutorials we've developed. One of the main issues was how to get the data, so this adds a ``require_data`` function that can be used to fetch files from the ``glue-example-data`` repo. I took the approach of simply downloading the data and not auto-loading to make sure that users see how to properly load a data file from disk.

I've also added a validation script that checks that the notebooks run without any issues.


astrofrog:
The travis failure is a fluke due to a timeout, so I'll go ahead and merge this. If the timeouts happen too often, I'll find a way to address it.

PR #68, opened 2019-04-04
Clean up viewer code

astrofrog:
I'm going through the code and simplifying things that can now be simplified due to updates in the core package. I've also organized the bqplot viewer code into sub-directories as is cleaner that way and mirrors what is done in the glue-core package. I plan to continue this cleanup, but I will do it over several PRs to keep diffs reasonable.


astrofrog:
Ok it looks like this requires the latest developer version of glue, so I'll need to update the CI

astrofrog:
@maartenbreddels - indeed there are no logic changes in this PR, just moving some code around. I'll make sure CI is passing before merging (I need to backport some glue dev improvements to get things to pass, but I'm isolating those in a ``compat.py`` file)

Issue #65, opened 2019-02-21
Strange selection issue in scatter viewer


astrofrog:
When playing around with the GAIA example at:

https://mybinder.org/v2/gh/glue-viz/glue-jupyter/master?filepath=notebooks%2FAstronomy%2FGAIA%2FDistance%20to%20The%20Pleiades%20with%20Glupyter%20and%20Gaia%20DR2.ipynb

I ran into a weird selection issue with the rectangular selection:

![out](https://user-images.githubusercontent.com/314716/53207647-2e043780-362c-11e9-862c-1600984ecbe2.gif)

PR #69, opened 2019-04-04
Fix bug that caused rectangular selection to not work properly

astrofrog:
Fixes https://github.com/glue-viz/glue-jupyter/issues/65


astrofrog:
@maartenbreddels - the test that was there was ambiguous because x2 == y1, so I've fixed that now.

PR #72, opened 2019-04-05
Generalize combo helper

astrofrog:
This generalizes the ``Component`` helper to not be specific to components but any SelectionCallbackProperty (it already wasn't specific apart from terminology) and also makes use of it wherever possible.

However, at the moment, the UI is a bit inconsistent as a result:

Screenshot 2019-04-05 at 18 10 41

@maartenbreddels - is there any way to at least make it so that the label and the dropdown are on the same line with ipymaterialui? Or is that not consistent with the material UI design?


astrofrog:
@maartenbreddels - I've just added a LinkedDropdown that uses the original widgets UI and also auto-updates the choices if they are changed (a feature I haven't yet added to the material UI one).

I think that for now (this PR) I'd like to switch back (for the dropdowns) to using the ipywidgets UI, and then we can switch over all in one go to a material UI if we want (but I think we should avoid mixing for now). Does that seem reasonable?

While I was working on this, I was thinking it would be much simpler if we could have a material library that exposed basically the same API as ipywidgets (and it might also make it more usable by others as a drop-in replacement)

maartenbreddels:
> think that for now (this PR) I'd like to switch back (for the dropdowns) to using the ipywidgets UI

To be honoust, I do not know why I switched that component, I think only the selection and icons were an issue. I agree, it will look more consistent if you take the classical.

I am not sure we'll have a 'classical' widgets interface, depends a bit how much work that is, but it would certainly easy the pain a bit to move from one to the other.

astrofrog:
I'll go ahead and merge this so that UI is consistent for now and also so the dropdown correctly updates if the number of choices changes. I've kept the material code too though so it's easy to switch.

Issue #19, opened 2018-08-26
Support >2 dimensional data in image viewer


astrofrog:
It would be nice to be able to show cubes in the image viewer. This will require including a tab for viewer state and then including sliders as needed, as for the Qt version.

pllim:
This was what I needed for spacetelescope/dat_pyinthesky#28

astrofrog:
Done in #75

PR #74, opened 2019-04-09
Added slicing capabilities to image viewer

astrofrog:
This also fixes a bug in LinkedDropdown related to updating options - I should add a regression test for this.

Fixes https://github.com/glue-viz/glue-jupyter/issues/19

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ install:
   - pip install -e .[test]
 
 script:
-  - pytest --cov glue_jupyter glue_jupyter
+  - pytest --cov glue_jupyter glue_jupyter -p no:warnings
   - pip install astroquery pyyaml
   - python .validate-notebooks.py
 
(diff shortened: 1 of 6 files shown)


pllim:
```python
from glue_jupyter.data import require_data
require_data('Astronomy/L1448/l1448_13co.fits')

import glue_jupyter as gj
app = gj.jglue()

cube = app.load_data('l1448_13co.fits')

app.imshow(data=cube)
```

It didn't crash but I don't see the data inside the plot. Also I don't know how to tell it which slice I want.

Screenshot from 2019-04-10 15-57-29

astrofrog:
@pllim - does it never appear even if you try multiple times? I've run into an issue where sometimes the image doesn't appear but sometimes it does, and I haven't quite tracked it down yet.

For the slice, you can select it using the Vopt slider on the right.

pllim:
FWIW I didn't try to run the same cell again and again. I did run the notebook multiple times over different sessions and had zero success.

astrofrog:
Ok no problem, I'll investigate - thanks!

I'm going to close this in favor of #75 since we might as well just merge it all in one go

PR #75, opened 2019-04-10
Add slicing, on-the-fly reprojection, and RGB compositing to image viewer, refactor state widgets

astrofrog:
The aim of this PR is to switch the image artists to ones similar to the ones in glue to handle the image RGB compositing as well as the reprojection. While doing this, I realized that we should also do a better job of separating out the widgets that relate to updating state objects from the viewers themselves, so that those widgets can be re-used with other viewers, such as the Matplotlib-based viewers. So I've made (for the image viewer) ``ImageLayerStateWidget``, ``ImageSubsetLayerStateWidget``, and ``ImageViewerStateWidget`` which inherit from VBox.

This might end up being quite a large diff, but most of this (apart from the updates to the image viewer) is moving code around.

@pllim - could you take this for a spin to see if it resolves your issue with the image not appearing?


astrofrog:
Ok so this should be ready to go - there are still places where we can clean up some things, make all the class names more consistent, etc. but I think this basically has everything working as it did before but the code is now organized in a more modular way.

pllim:
> could you take this for a spin

Sounds like Sara wants me to switch gears, so I don't think I can do it now, despite what I said on the phone 10 mins ago. 😬

glue_jupyter/bqplot/histogram/viewer.py
```diff
@@ -1,54 +1,26 @@
...
-
-        # self.widget_hist_n_bin = widgets.IntSlider(min=1, max=1000, step=1, description='bins')
-        # link((self.state, 'hist_n_bin'), (self.widget_hist_n_bin, 'value'))
-
-        self.tab_general.children += (widgets.HBox([self.button_normalize, self.button_cumulative] ),) + tuple(self.widgets_axis)
+    _state_cls = HistogramViewerState
+    _options_cls = HistogramViewerStateWidget
+    _data_artist_cls = BqplotHistogramLayerArtist
+    _subset_artist_cls = BqplotHistogramLayerArtist
+    _layer_style_widget_cls = HistogramLayerStateWidget
```
maartenbreddels: good idea!

glue_jupyter/bqplot/image/frb_mark.py
```diff
@@ -0,0 +1,68 @@
+# This is an image sub-class that automatically gets its values from an object
+# that implements __call__ with a bounds= argument. It is the equivalent of
+# FRBArtist in glue-core.
```
maartenbreddels: What does FRB stand for?

glue_jupyter/bqplot/image/layer_artist.py
```diff
@@ -1,182 +1,82 @@
...
+    _layer_state_cls = ImageSubsetLayerState
+
+    def __init__(self, axes, viewer_state, layer_state=None, layer=None):
+
+        super(BqplotImageSubsetLayerArtist, self).__init__(axes, viewer_state,
+                                                           layer_state=layer_state, layer=layer)
+
+        self.subset_array = ImageSubsetArray(self._viewer_state, self)
+
+        self.image_artist = FRBImage(axes, self.subset_array)
```
maartenbreddels: A bit confused here, the first argument is axes, which is passed along here, but in FRBImage it is viewer, shouldn't this be viewer? (copy paste mistake from matplotlib code maybe?)

glue_jupyter/bqplot/scatter/viewer_options_widget.py
```diff
@@ -0,0 +1,17 @@
...
+class ScatterViewerStateWidget(VBox):
+
+    def __init__(self, viewer_state):
+
+        self.state = viewer_state
+
+        self.widget_x_axis = LinkedDropdown(self.state, 'x_att', label='x axis')
+        self.widget_y_axis = LinkedDropdown(self.state, 'y_att', label='y axis')
+
+        super().__init__([self.widget_x_axis, self.widget_y_axis])
```
maartenbreddels: this super() means we drop python2 support, did we already agree (I do btw)

glue_jupyter/link.py
```diff
@@ -10,12 +10,14 @@ def _is_echo(link):
...
+    def __init__(self, source, target, f1=lambda x: x, f2=lambda x: x, forwards=True, backwards=True):
         self.source = source
         self.target = target
 
-        self._link(source, target, 'source', f1, True)
-        self._link(target, source, 'target', f2)
+        if forwards:
+            self._link(source, target, 'source', f1, True)
+        if backwards:
+            self._link(target, source, 'target', f2)
```
maartenbreddels: we could also use dlink (which traitlets and ipywidgets has as well), to keep the api similar. If you prefer this i'm ok with it. I wonder about the last argument (sync_directly), when forwards is False, they will not be synced.

glue_jupyter/bqplot/image/frb_mark.py
astrofrog: Fixed resolution buffer - I'll write it out in the comments

glue_jupyter/bqplot/image/layer_artist.py
astrofrog: Yes it was from copy/pasting - will fix

glue_jupyter/bqplot/scatter/viewer_options_widget.py
astrofrog: Yes we already did in https://github.com/glue-viz/glue-jupyter/pull/57

glue_jupyter/link.py
astrofrog: Oh I see, I hadn't realized dlink does what we need already - will fix.

PR #78, opened 2019-04-12
Add a bqplot profile viewer

astrofrog:
We already had a simple Matplotlib profile viewer - this adds the bqplot version.


glue_jupyter/bqplot/profile/layer_artist.py
```diff
@@ -0,0 +1,162 @@
...
+        self.line_mark.colors = [self.state.color]
+        self.line_mark.opacities = [self.state.alpha]
+
+    def _calculate_profile(self, reset=False):
+        try:
+            self._calculate_profile_thread(reset=reset)
+        except Exception:
+            self._calculate_profile_error(sys.exc_info())
+        else:
+            self._calculate_profile_postthread()
```
astrofrog: The mention of threads is because this is adapted from the code in glue-core, which is multi-threaded. I'm trying to keep the code as similar as possible to the core code so that we can later try and reduce duplication.

Issue #82, opened 2019-04-13
Add an option to change the behavior of the debounce function


astrofrog:
When panning around, the current debounce method works nicely. However, when changing slice in the n-dimensional data, it would be nice to have the image change while the slice is being changed, but at the same time we don't want all the calls to build up, so we still want something similar to debounce, but which does guarantee execution of the function as fast as possible, dropping any calls that pile up.

maartenbreddels:
So would you want in these terms:
https://css-tricks.com/debouncing-throttling-explained-examples/
a 'leading debouce'?

PR #86, opened 2019-04-13
Add a mode to debounced to start executing the function while events are still being received

astrofrog:
This is an attempt to solve https://github.com/glue-viz/glue-jupyter/issues/82, but not quite there yet. For some background, the debounced decorator works fine in master when panning/zooming - the image only updates if the user stops panning for 0.5 seconds. However, for changing e.g. slices, or bias/contrast, I'd like to start getting feedback on the change while I'm dragging the slider, not just when I stop.

I want the calls to happen as fast as possible but at the same time I don't want the image update calls to build up if they are a bit slow, so I'm trying to use/modify ``debounced`` so that the function is guaranteed to execute at regular intervals. The approach here kind of works but for some reason the updates separated by quite a long delay while dragging the slider, and then when I let go, there is still a backlog of calls, so not quite there yet.

To put this a different way, let's say that the image update takes 1 second, I ideally want to see the image update happen every second, and overall, the last update should happen <=1 second from the last event.

@maartenbreddels - does this make sense? Any ideas on how to make this work properly?

### DIFF
--- a/glue_jupyter/bqplot/image/frb_mark.py
+++ b/glue_jupyter/bqplot/image/frb_mark.py
@@ -33,7 +33,7 @@ class FRBImage(AstroImage):
         self.viewer.figure.axes[1].scale.observe(self.update, 'min')
         self.viewer.figure.axes[1].scale.observe(self.update, 'max')
 
-    @debounced(method=True)
+    @debounced(method=True, wait_for_end=False, delay_seconds=0.1)
     def update(self, *args, **kwargs):
 
         # Get current limits from the plot
--- a/glue_jupyter/bqplot/image/frb_mark.py
+++ b/glue_jupyter/bqplot/image/frb_mark.py
@@ -50,16 +50,16 @@ class FRBImage(AstroImage):
         # make it a parameter in future. Bqplot does allow us to constrain the
         # aspect ratio though so we could envisage trying to use that info
         # to make sure the ratio of the sizes is sensible..
-        ny, nx = 512, 512
+        ny, nx = 256, 256
 
         # Set up bounds
         bounds = [(ymin, ymax, ny), (xmin, xmax, nx)]
 
         # Get the array and assign it to the artist
+        image = self.array_maker(bounds=bounds)
+        if image is None:
+            image = np.zeros((1, 1, 3))
         with self.hold_sync():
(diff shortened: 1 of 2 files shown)


maartenbreddels:
Maybe throttling is a better idea/word:
https://css-tricks.com/debouncing-throttling-explained-examples/#article-header-id-5
Sometimes it's difficult to actually put in words what you want, let alone code.

astrofrog:
Yes, basically what I need is throttling, but rather than having a specific rate of throttling, the rate should essentially be set by the execution time of the function - so that basically we run the function as many times as possible without building up a backlog.

The following example illustrates what I mean with a thread:

```
from threading import Thread
from queue import Queue, Empty

def queue_to_list(q):
    
    # Start off by waiting for an object to be added to the queue
    l = [q.get()]
    
    # Now check if there are any other items in the queue, which we can just do asynchronously
    while True:
        try:
            l.append(q.get_nowait())
        except Empty:
            return l
    

class CallbackHandler(Thread):

    def __init__(self, function):
        super().__init__()
        self.function = function
        self.work_queue = Queue()

    def run(self):

        while True:

            msgs = queue_to_list(self.work_queue)
                    
            if 'stop' in msgs:
                return
            elif len(msgs) > 0:
                area2.value += f'Combining {len(msgs)} messages\n'
                self.function()
                
handler = CallbackHandler(callback)
handler.start()

def add_to_queue(*args):
    area1.value += "Adding to queue\n"
    handler.work_queue.put('run')
    
slider.observe(add_to_queue, 'value')
```

In fact, having thought about this some more, I think threads might indeed be needed for this kind of behavior. I'll see if I can write it up as a decorator.

astrofrog:
Ok so I managed to implement a decorator version of what I need (to opt in to that, the decorator in ``frb_mark.py`` should be changed from ``@debounced`` to ``@thread_throttled(delay=0.1)``. However, there are still issues with the image updating quite slowly while the slider is being changed, and then continuing to update, faster, once the slider stops moving. I think what's happening is actually that the bottleneck must be in the communication between JS and Python. At this point, I'm not sure what the best way is around this.

astrofrog:
I haven't really noticed issues related to this recently, so am just going to close this and consider re-opening if it looks like we need this kind of solution.

PR #89, opened 2019-04-14
Move selection buttons in bqplot into toolbar above viewer

astrofrog:
Just a minor UI change, I think it would make more sense for the selection tools to be next to the rest of the selection toolbar:

Screenshot 2019-04-14 at 17 15 40


maartenbreddels:
Yes, makes total sense, this grew organically since there was no such toolbar first

PR #115, opened 2019-06-07
Started re-implementing layout with vuetify

astrofrog:
This starts to implement a vuetify-based interface for viewers. To try it out:

```python
from glue_jupyter import set_layout_factory
from glue_jupyter.vuetify_layout import vuetify_layout_factory
set_layout_factory(vuetify_layout_factory)
```

Preview of scatter viewer with sidebar collapsed:

Screenshot from 2019-06-07 17-43-22

and with sidebar open:

Screenshot from 2019-06-07 17-43-32

I haven't yet re-implemented the way to allow multiple subset selections, but it's definitely possible. Also, the viewer and layer options in the sidebar still use ipywidgets - for now we might want to keep this as is since it's actually pretty compact.

### DIFF
--- a/glue_jupyter/common/toolbar_vuetify.py
+++ b/glue_jupyter/common/toolbar_vuetify.py
@@ -18,7 +18,6 @@ class BasicJupyterToolbar(v.BtnToggle):
         self.output = viewer.output_widget
 
     def _change_tool(self, widget, event, data):
-
         with self.output:
 
             if self.active is not None:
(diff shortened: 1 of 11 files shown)


glue_jupyter/common/toolbar_vuetify.py
```diff
@@ -0,0 +1,36 @@
...
+
+__all__ = ['BasicJupyterToolbar']
+
+ICON_WIDTH = 20
+
+
+class BasicJupyterToolbar(v.BtnToggle):
+
+    def __init__(self, viewer):
+        super().__init__(v_model='toggle_exclusive', class_='transparent')
```
maartenbreddels: v_model can be a string, but it will be set to the value of the selected button, by default it exclusive.

maartenbreddels:
I added a colored triangle to the subset selection:
image

A circle was confusing since there is a red circle to the left as well. But the usability is better than no color I think.

astrofrog:
I've now improved (hopefully) the sidebar:

Screenshot 2019-09-17 at 16 45 55


and made vuetify the default layout. It's still possible to opt in to the old layout with:

```
from glue_jupyter import set_layout_factory
from glue_jupyter.ipywidgets_layout import ipywidgets_layout_factory
set_layout_factory(ipywidgets_layout_factory)
```

PR #124, opened 2019-08-12
Replace deprecated Div reference with Html

nmearl:
The `Div` bindings seems to have been removed in the latest versions of ipymaterialui. Replacing with the `Html` component seems to solve the issue.


astrofrog:
Thanks!

PR #125, opened 2019-08-14
Explicitly set toolbar children if uninitialized

nmearl:
Apparently it's possible that an ipywidget's children attribute is not initialized as an empty list object. In that case, check first before concatenating lists.


astrofrog:
Thanks! But closing in favor of https://github.com/glue-viz/glue-jupyter/pull/126 since that includes all the fixes required to get things working again.

PR #132, opened 2019-09-24
feat: update to ipyvuetify 1.0 (Vuetify 2.0)

mariobuikhuizen:


glue_jupyter/vuetify_layout.py
```diff
@@ -19,18 +19,22 @@ def vuetify_layout_factory(viewer):
...
+                v.ExpansionPanelHeader(class_='font-weight-bold', children=['Viewer Options']),
+                v.ExpansionPanelContent(children=[viewer.viewer_options])]),
+            v.ExpansionPanel(children=[
+                v.ExpansionPanelHeader(class_='font-weight-bold', children=['Layer Options']),
+                v.ExpansionPanelContent(children=[viewer.layer_options])])])
 
     drawer = v.NavigationDrawer(v_model=False, absolute=True, right=True,
                                 children=[sidebar_button,
-                                          options_panel], width=350)
+                                          options_panel], width=373)
```
maartenbreddels: why this magic number? should we add a comment in the code?

maartenbreddels:
> Perhaps @astrofrog knows a bit more, but it seems like the selection behavior is active _even for that single click outside the region_.

I checked, but this behavior is already in master.



> Also, I know this isn't glue-jupyter specific, but I'd like to say that there is a certain awkwardness to the scoped slots behavior... having two children lists feels weird to me.

It's ipyvue specific. I agree that is awkward, but I have to say that scoped slots are awkward by itself, and putting them as template-children (at the same level as the content of e.g. v-menu) is also quite unintuitive.

Our reasoning was the following. Either we completely model it like vue, so scoped slots would be children, wrapped by a `v.Template(..)`, but that is quite verbose. Also, the only reason I can think of having scoped slots as children in a vue template, is because there is no other natural way in xml/html syntax, while in Python there would be other ways.

The way we implemented scoped slots in ipvue 1.0, is (we think) more Pythonic. But, if you have better ideas we are all ears, if we can make it even easier, that would be a win for all.

Given that this PR doesn't change behaviour, and upgrades only to ipyvuetify1.0/vuetify2, I think we can merge this.

PR #141, opened 2020-01-27
Fix case where Data.coords is None

astrofrog:
This fixes the test suite for the changes in https://github.com/glue-viz/glue/pull/2079 - but in fact the changes here are more general, and are to make sure things work correctly if ``Data.coords`` is ``None``


astrofrog:
Merging since this is a trivial fix

PR #142, opened 2020-01-29
Generalize subset-related global widgets to work with session objects not just viewers

astrofrog:
cc @nmearl


astrofrog:
This is reasonably straightforward so I'll go ahead and merge

PR #143, opened 2020-01-29
Remove ipymaterialui-related code

astrofrog:
@nmearl made me aware of an issue whereby ``viewer.toolbar`` would always return the materialui toolbar, not e.g. the vuetify one, and similarly for other viewer attributes. This PR removes all materialui-related code to make everything consistent again. Some tests need to be updated for the CI to pass, so this is a WIP.

@maartenbreddels - is there any reason to keep the materialui code around?


astrofrog:
I'm going to go ahead and merge this once tests pass since it's needed to fix issues in jdaviz. If we decide to ever add back support for ipymaterialui, we can dig the widgets out of the history, but for now I think it'll be easier to not have two parallel sets of widgets to maintain.

Issue #144, opened 2020-02-09
Change default marker used for 3-d scatter


astrofrog:
The default marker used for the 3d scatter (ipyvolume) viewer is currently 'sphere', which is the slowest. We should change this to 'box' (and make sure it is indeed faster).

astrofrog:
@jsub1 - would you be happy to look into this? It should be a very easy fix.

maartenbreddels:
What about circle_2d? It has less 3d effect, but should be faster.

(from mobile phone)

astrofrog:
I haven't tried, but that sounds promising - it would be similar to what we do in the Qt interface

jsub1:
The diamond marker renders faster than either the box or circle_2d (23.66 fps with 1.5M points vs 18.02 or 14.81). We should default to the diamond unless there are aesthetic reasons not to.

astrofrog:
Sounds good to me!

maartenbreddels:
Can I ask how you measured the FPS?

astrofrog:
I'll close this but I'd be interested in also knowing how the FPS was measured!

jsub1:
Sorry about the delay in answering this. 

What I did was to create a small looping animation using the movie maker. I then used the performance section of the developers tool and recorded multiple loops of the short animation.

 The numbers quoted come from Firefox, because that performance tool will just display the average, min and max FPSs unlike Chrome. The numbers above are average.

PR #149, opened 2020-02-11
Scatter 3D Default Marker

jsub1:
Fix for #144. Change the default scatter3d marker to diamond for performance reasons. Also add circle_2d as a marker option for a better performing circular option.


astrofrog:
Before I merge, I noticed your commit doesn't have your avatar, which indicates that you are using a different email address for your git config than for github - is that intentional?

jsub1:
@astrofrog It was set to my old work email. I've changed it now.

maartenbreddels:
LGTM, although may be useful for ipyvolume to expose/export which markers are possible?

@jsub1 if you edit the authorship of your commit you get credits for it (you need to force push for that).

PR #152, opened 2020-02-18
Convert layer and viewer options widgets to use ipyvuetify

astrofrog:
This is a work in progress to convert the viewer and layer options from ipywidgets to ipyvuetify widgets. For now, only the histogram viewer has been converted. I've now tried to make full use of vuetify yet, rather focusing on replicating a UI similar to the ipywidgets one and connecting everything to the glue states. The layout is currently not very compact:

Screenshot from 2020-02-18 15-47-04
Screenshot from 2020-02-18 15-47-11

@mariobuikhuizen @maartenbreddels - it would be great if you could take a stab at or give advice for how we can make this layout more compact. For the color picker, I actually quite like the ipywidgets version of having a pop up dialog to change the color as opposed to having it be always visible, but I'm guessing this may not be possible with ipyvuetify? To try out the current PR, you can use the following in a notebook:

```python
import numpy as np
import glue_jupyter as gj
app = gj.jglue(data=np.random.normal(0, 1, 100))
histogram_viewer = app.histogram1d()
```


astrofrog:
For the 'visible' checkbox I'm wondering whether it would be possible to have checkboxes in the dropdown of layers itself which would make it easier to enable/disable many layers in one go. Is that possible?

mariobuikhuizen:
I did some experimenting: https://codepen.io/mariobuikhuizen/pen/XWbKVKM. It can get pretty dense!

Schermafbeelding 2020-02-18 om 23 05 13

mariobuikhuizen:
> For the 'visible' checkbox I'm wondering whether it would be possible to have checkboxes in the dropdown of layers itself which would make it easier to enable/disable many layers in one go. Is that possible?

Something like this? https://codepen.io/mariobuikhuizen/pen/YzXGKaP
Schermafbeelding 2020-02-19 om 17 47 11

Schermafbeelding 2020-02-19 om 17 47 26

astrofrog:
Looks nice!

mariobuikhuizen:
When I try to use `GlueState()` I get an error:
```
     51 
     52     def convert_state_to_json(self, state, widget):
---> 53         return json.dumps(state.as_dict(), cls=GlueStateJSONEncoder)
     54 
     55     def update_state_from_json(self, json, widget):

AttributeError: 'NoneType' object has no attribute 'as_dict'
```

Code to reproduce:
```
class HistogramViewerStateWidget(v.VuetifyTemplate):
    ...
    glue_state = GlueState().tag(sync=True)

    def __init__(self, viewer_state):
        super().__init__()
        glue_state = viewer_state
        ...
```

astrofrog:
@mariobuikhuizen - can you try again now?

mariobuikhuizen:
@astrofrog It works! See https://github.com/astrofrog/glue-jupyter/pull/6.

The new features in ipyvue and ipyvuetify this PR depends on are released.

mariobuikhuizen:
@astrofrog I've made a PR for the refactoring of the profile and image viewer: https://github.com/astrofrog/glue-jupyter/pull/7.

What do you think?

astrofrog:
@mariobuikhuizen - looks great, thanks! Just going to push a few tweaks then merge this in, and then do follow-up PRs for any small issues.

astrofrog:
Once the CI passes, I'll go ahead and merge this since it works nicely, and we can do follow up PRs to tweak things.

PR #163, opened 2020-03-05
Make the 'allow multiple subsets' toggle optional and disable it by default

### LINKED ISSUE #159: Make it possible to enable/disable the allow multiple subsets option
The subset selection menu has an 'allow multiple subsets' option which is pretty advanced. We should hide it by default and provide an easy way to enable it in an 'advanced' mode.

astrofrog:
For now enabling this option will require modifying a trait on the widget, which can't easily be done via the notebook, but in future we could choose to expose an 'advanced' mode that would enable this (and other) options.

Fixes #159

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,9 @@
 * Removed ipymaterialui widgets and fix cases where these widgets were
   used over ipyvuetify widgets. [#143]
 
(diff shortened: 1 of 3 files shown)


astrofrog:
@mariobuikhuizen - does this seem ok?

mariobuikhuizen:
Yes, perfect.

PR #166, opened 2020-03-09
Fix removal of data from bqplot viewers

astrofrog:
@nmearl - this should fix the removal of data from bqplot viewers

### DIFF
--- a/glue_jupyter/bqplot/histogram/layer_artist.py
+++ b/glue_jupyter/bqplot/histogram/layer_artist.py
@@ -36,8 +36,10 @@ class BqplotHistogramLayerArtist(LayerArtist):
         link((self.state, 'visible'), (self.bars, 'visible'))
 
     def remove(self):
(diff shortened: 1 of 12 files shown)


nmearl:
Thanks @astrofrog! However, while the error is gone, and the "Layer" field in the layer options widget is empty, the actually plotted data is not removed from the viewer.

glue_jupyter/bqplot/histogram/layer_artist.py
```diff
@@ -35,6 +35,13 @@ def __init__(self, view, viewer_state, layer_state=None, layer=None):
 
         link((self.state, 'visible'), (self.bars, 'visible'))
 
+    def remove(self):
+        marks = self.view.figure.marks[:]
+        marks.remove(self.bars)
+        self.bars = None
+        self.view.figure.marks = marks
```
astrofrog: @maartenbreddels - is there a better way to remove marks from a bqplot figure?

PR #170, opened 2020-03-20
Fix behavior of selection tools

astrofrog:
This reverts a change I made in #164 which de-selected a selection tool after a selection was made.

There are still some issues with some events being still passed to the circular selection even once the circular selection is deselected. See the following screencast:

Peek 2020-03-20 15-56

This shows:

* Making rectangular selections works
* Once I deselect the tool selections don't work (=good)
* When I select the circular selection tool selections work
* When I switch back to the rectangular selection there is a small glitch
* When I disable the rectangular selection now clicking causes the circular selection to come back and get updated.
* Another issue is that the circular selection brush color seems to be ignored and is shown in green

@maartenbreddels - could you see if you can reproduce this and investigate what could be going wrong with the circular brush?

### DIFF
--- a/glue_jupyter/bqplot/common/tools.py
+++ b/glue_jupyter/bqplot/common/tools.py
@@ -62,6 +62,8 @@ class BqplotRectangleMode(InteractCheckableTool):
             self.update_from_roi(roi)
 
         self.interact.observe(self.update_selection, "brushing")
+        self.interact.observe(self.on_selection_change, "selected_x")
+        self.interact.observe(self.on_selection_change, "selected_y")
         self.finalize_callback = finalize_callback
 
     def update_selection(self, *args):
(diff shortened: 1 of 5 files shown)


maartenbreddels:
The issue for the 

> * When I switch back to the rectangular selection there is a small glitch

I wonder if the glitch is in glue? Or that it's fixed similar to the patch I propose below. The reason why I think it might be in glue, is that before this PR, if you select rectangles multiple times, it selects a larger region than what is selected by the rectangle. However inspecting `t = image_viewer.toolbar.tools['bqplot:rectangle'].interact.selected` shows the right coordinates.
For some reason though, this bug does not show up (as often?) in this branch.






> * When I disable the rectangular selection now clicking causes the circular selection to come back and get updated.

Fixed by https://github.com/glue-viz/bqplot-image-gl/pull/18



> * When I disable the rectangular selection now clicking causes the circular selection to come back and get updated.

That could be considered a bug in the other interacts, or a feature :). We could have the same behaviour as the rectangle brush by doing this:
```patch

diff --git a/glue_jupyter/bqplot/common/tools.py b/glue_jupyter/bqplot/common/tools.py
index 20bbf19..2134993 100644
--- a/glue_jupyter/bqplot/common/tools.py
+++ b/glue_jupyter/bqplot/common/tools.py
@@ -86,6 +86,12 @@ class BqplotCircleMode(InteractCheckableTool):
 
         self.interact.observe(self.update_selection, "brushing")
 
+    def activate(self):
+        with self.viewer._output_widget:
+            self.interact.selected_x = None
+            self.interact.selected_y = None
+        super().activate()
+
     def update_selection(self, *args):
         with self.viewer._output_widget:
             if self.interact.selected_x is not None and self.interact.selected_y is not None:

```

astrofrog:
@maartenbreddels - thanks for investigating this, for fixing bqplot-image-gl, and for the other suggested fix! (which indeed also fixed the glitch I mentioned). Regarding the color, I have a workaround for now.

PR #174, opened 2020-04-02
Load plugins when calling JupyterApplication

astrofrog:
@nmearl - this should ensure that glue plugins are loaded in jdaviz, including those required for the auto-linking.


nmearl:
Ah, right, I totally forgot about this. 😅

PR #176, opened 2020-04-07
Display proper units information on bqplot axes

dtn5ah:
This small change displays the associated world attribute units on the bqplot axes.


dtn5ah:
Looks like I made some false assumptions about available state attributes. Pulling back this PR

PR #182, opened 2020-05-04
Fix syncing of state from JSON when nested MAGIC_IGNORE values were present

astrofrog:
@nmearl - this should fix the issue you were seeing with UUIDs making their way back to the glue state.


nmearl:
@astrofrog yep, this fixes the issues I was seeing locally! Thanks!

PR #200, opened 2020-08-31
fix: image viewer lags behind trying to render each slider position

mariobuikhuizen:
Using lodash the updates are limited to 1 each 300ms


rosteen (approved):
Looks good to me - I tested with jdaviz, the slider behavior definitely seems better when scrolling quickly through data.

Edit: Actually, `jdaviz` wasn't set up to actually use this slider like I thought - I'll have to re-test.

astrofrog:
@mariobuikhuizen - thanks! Is there a reason for choosing 300ms? (which seems quite slow)

mariobuikhuizen:
That value was suggested by the reporter of the jdaviz issue. What's a good value will depend on the situation, so maybe this should be configurable. Ultimately something like @dtn5ah describes in https://github.com/spacetelescope/jdaviz/issues/244#issuecomment-687452279 would be nice.

PR #212, opened 2021-03-23
refactor: use VuetifyTemplate to reduce the number of widgets

### LINKED ISSUE #209: Performance: turn SelectionModeMenu into a template
SelectionModeMenu is made up of quite some widget, and therefore takes up quite some time (every `open` is opening a Comm object, which belongs to 1 widget):

image

I think we should turn this into a template

mariobuikhuizen:
This shortens Voila load times.

Fixes #209


maartenbreddels:
This is excellent, I see speedups in (pure) glue-jupyter of up to 2x for creating a scatter viewer (~500ms to ~250ms) with this. I think we can do the same for the subsetmode tool (add/remove/subtract/xor)

PR #213, opened 2021-03-29
Start implementing 'always-on' MouseInteraction

astrofrog:
@maartenbreddels - this is what I have so far but I have run into an issue which is that the print statement to show when mouse interactions are received only works if you first select a tool, e.g. rectangular selection. For some reason it doesn't work before this, when ``next`` is set to ``None``.

Note that there is also a default 'click and drag ROI' mouse mode that gets selected when other tools aren't selected - this doesn't seem to play nicely with having a MouseInteration that has a ``next `` which is the ROI ``MouseInteraction``, so this will also require some investigation.

Note that this requires https://github.com/glue-viz/bqplot-image-gl/pull/46


maartenbreddels:
@astrofrog would be great if you can try again with https://github.com/glue-viz/bqplot-image-gl/pull/47

astrofrog:
@maartenbreddels - the CI is being a bit slow to start but this is ready for review.

astrofrog:
(I just rebased this)

astrofrog:
This seems to work fine for https://github.com/spacetelescope/jdaviz/pull/429

maartenbreddels (approved):
Looks good, the only 'downside' i see is that we'll always get a lot of events. We could possibly add an extra argument to the `add_event_callback` which events we want to listen to, and set `_mouse_interact.events` based on that. But, we can postpone that for now.

astrofrog:
@maartenbreddels - yes good idea. I'll merge this for now and will add something like this in a follow-up PR.

pllim:
Is this released? If not what is the timeline? This is blocking spacetelescope/jdaviz#429

PR #214, opened 2021-03-31
Implement JupyterApplication.viewers

astrofrog:
Prior to this PR this defaulted to the base class implementation which was an empty list


astrofrog:
Merging this as it should be uncontroversial

PR #216, opened 2021-04-21
MNT: Remove print colormap

pllim:
# Description

I suspect this is the source of the bug seen in spacetelescope/jdaviz#548


pllim:
This was added in #152

astrofrog (approved):
Oops!

PR #218, opened 2021-04-27
Added 'home' tool in bqplot viewers to reset limits

astrofrog:
This works as for the Matplotlib viewers and I think users should be familiar with the icon.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+0.4 (unreleased)
+================
+
+* Added a new 'home' tool in bqplot viewers to reset limits. [#218]
+
 0.3 (2021-04-15)
 ================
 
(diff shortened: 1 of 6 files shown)


rosteen:
I confirmed that this works, thanks! We've definitely had a bunch of users/testers requesting this feature. I'm not a big fan of the icon but I think we can override that in Jdaviz if we don't like it, certainly not a blocker here.

rosteen (approved):

astrofrog:
@rosteen - yes I'm happy to change the icon here too once we have a better option! (might as well be consistent)

pllim:
Thanks! Is there a planned release to include this and recent fixes, or you still waiting for something else?

rosteen:
For what it's worth, we mostly use the material design icons (which are included in ipyvuetify) in Jdaviz. Our UI/UX person designed some custom icons but I don't think we've actually switched to using any of them (whoops).

astrofrog:
@pllim I'll try and do a release on Thursday!

astrofrog:
@pllim - I managed to release v0.4 today: https://pypi.org/project/glue-jupyter/0.4/

Issue #220, opened 2021-04-29
BUG: Home disables pan/zoom for one turn


pllim:
**Describe the bug**

This is a direct follow-up of #218 added in `glue-jupyter` 0.4.

**To Reproduce**

See https://github.com/spacetelescope/jdaviz/pull/575#issuecomment-829314245 and https://github.com/spacetelescope/jdaviz/pull/575#issuecomment-829315449

**Expected behavior**

Pan/zoom would work normally even after I clicked Home.

**Screenshots**

N/A

**Details:**

- Operating System: RHEL 7
- Python version (``python --version``): [e.g. Python 3.6.7] 3.8.3
- Glue version (``glue --version``): [e.g. 0.14.5] 1.0.1
- How you installed glue: [e.g. conda or pip] pip

**Additional context**

Seen in `jdaviz` Imviz.

pllim:
I think this is fixed by #222

PR #222, opened 2021-05-06
Improvements to tools

astrofrog:
This does two things:

* It keeps better track of what callbacks we enable/disable when activating/deactivating tools
* It fixes the tools in the case where a single-shot (non-checkable) tool is used


pllim:
So this is supposed to fix glue-viz/glue-jupyter#220 ?

pllim:
@astrofrog , when you do another release with this fix included, please also bump `glue-jupyter` minversion over at `jdaviz`. Thanks!

PR #227, opened 2021-05-25
Use LinesGL for faster rendering of lines.

maartenbreddels:
It might be useful to switch off GL, using the `USE_GL` variable. Requires https://github.com/glue-viz/bqplot-image-gl/pull/58

### DIFF
--- a/glue_jupyter/bqplot/profile/layer_artist.py
+++ b/glue_jupyter/bqplot/profile/layer_artist.py
@@ -11,12 +11,16 @@ from glue.viewers.profile.state import ProfileLayerState
 from glue.core.exceptions import IncompatibleAttribute, IncompatibleDataException
 
 import bqplot
+from bqplot_image_gl import LinesGL
 
 from glue.viewers.common.layer_artist import LayerArtist
 
 from ...link import dlink
 
 __all__ = ['BqplotProfileLayerArtist']
+
+
+USE_GL = True
 
 
 class BqplotProfileLayerArtist(LayerArtist):
@@ -34,7 +38,8 @@ class BqplotProfileLayerArtist(LayerArtist):
 
         self.view = view
 
-        self.line_mark = bqplot.Lines(scales=self.view.scales, x=[0, 1], y=[0, 1])
+        LinesClass = LinesGL if USE_GL else bqplot.Lines
+        self.line_mark = LinesClass(scales=self.view.scales, x=[0, 1], y=[0, 1])
 
         self.view.figure.marks = list(self.view.figure.marks) + [self.line_mark]
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -22,7 +22,7 @@ install_requires =
     ipywidgets>=7.4.0
     ipyvue>=1.2.0,<2
     ipyvuetify>=1.2.0,<2
-    bqplot-image-gl>=1.4.1
+    bqplot-image-gl>=1.4.3
     bqplot>=0.12.17
     scikit-image
 


maartenbreddels:
Does the js console give a error?

rosteen:
Hmm, this is just in Logs not Errors/Warnings but I see: `THREE.WebGLRenderer 91 index.js:2:1246333`

Additionally, I saw this once, but it doesn't appear again if I restart the kernel and rerun:

```
Source map error: Error: request failed with status 404
Resource URL: http://localhost:8888/nbextensions/bqplot/index.js?v=20210525151615
Source Map URL: index.js.map
```

maartenbreddels:
Can you try out
https://github.com/glue-viz/bqplot-image-gl/blob/master/examples/lines.ipynb



(from mobile phone)

rosteen:
Hm, everything in that notebook works. Perhaps we're using `bqplot.Lines` explicitly somewhere in the Specviz code and that's getting messed up with LinesGL enabled...I'm actually remembering now that I have a `SpectralLine` that's subclassing `bqplot.Lines`, but that should mess up the basic display of the spectrum.

maartenbreddels:
https://github.com/glue-viz/bqplot-image-gl/pull/62 will fix this.

astrofrog:
@maartenbreddels - could you rebase this due to the small conflict?

@rosteen - could you test this again with the latest bqplot-image-gl release?

rosteen:
@maartenbreddels Sorry for the delay in getting back to you on this. I still see the same behavior in Specviz with the most recent `bqplot-image-gl` installed and this PR rebased to the current master branch - the Specviz app loads and data can be loaded into it, but no data is displayed on the plot. Still seeing very similar things in the console log as before (`THREE.WebGLRenderer 91 index.js:2:1247059` and some seemingly unrelated warnings).

Maybe we could screen share at some point to try to pin down what's going wrong.

rosteen:
I've been digging into this, and the best clue I've found is that switching from `World 0` to `Pixel Axis 0 [x]` for the `x-axis` in the viewer options causes the line to actually show up (using the World Axis works without this PR installed, as expected). I dug down a bit into the viewer attributes and it seems to think there's a line mark even in the case that the line is not appearing (this is with `World 0` as the `x-axis`):

```
>>> viewer.layers[0].line_mark

LinesGL(colors=['#595959'], interactions={'hover': 'tooltip'}, opacities=[1.0], scales={'x': LinearScale(allow_padding=False, 
max=9198.140625, min=3799.2685546875), 'y': LinearScale(max=168.5687698364258, min=12.555421447753906)}, 
scales_metadata={'x': {'orientation': 'horizontal', 'dimension': 'x'}, 'y': {'orientation': 'vertical', 'dimension': 'y'}, 'color': 
{'dimension': 'color'}}, stroke_width=1.0, tooltip_style={'opacity': 0.9}, x=array([3799.26855469, 3800.14257812, 
3801.01879883, ..., 9193.90527344, 9196.02050781, 9198.140625  ]), y=array([30.59662628, 33.24572754, 
35.89511871, ..., 53.27968979, 50.23616791, 51.70271683]))
```

I'll also note that I double checked that my `pip list` is identical other than the `glue-jupyter` version between the working case and the non-working case (this PR). 

As noted before, the spectrum generated by collapsing the data in Cubeviz successfully plots in its spectrum viewer, and if I load the Specviz example data after that it successfully displays. It seems that using that particular data as the reference data is causing the problem, but I still haven't figured out exactly why, given that it's always worked in the past.

maartenbreddels:
Just to check, are you running bqplot-image-gl 1.4.2 which includes https://github.com/glue-viz/bqplot-image-gl/commit/7044ad29afaa38914a993705627a29b15bfcd99e ?

rosteen:
I am indeed.

maartenbreddels:
https://github.com/glue-viz/bqplot-image-gl/pull/64 should fix this and what you describe makes sense give that depending in the scale of the axes, it will make threejs think it's inside/outside the viewport.

maartenbreddels:
@rosteen this branch should work now, I've tested it in a new environment, but I'd like to hear that it is also working for you!

rosteen:
Works for me now!

astrofrog (changes requested):
Looks good! However could you update this to include LinesGL in the new compat module you added in #302 for the other GL marks?

maartenbreddels:
This one is in bqplot_image_gl so it doesn't depend on bqplot 0.12/0.13 and bqplot-gl, so I think this is good as it is.

astrofrog (approved):
Ok sounds good!

PR #230, opened 2021-05-25
Remove layer options that aren't relevant to profile viewers

rosteen:
The `vmin`, `vmax` and `percentile` options are currently included in the profile viewer `Layer` options, but they aren't editable by the user and don't really do anything in the profile viewer. They seem to be related to color scaling in the 2D viewers and thus aren't relevant to the profile viewer, so I removed them.


maartenbreddels:
I agree with this PR, but I'd like @astrofrog 's opinion.

astrofrog (approved):

PR #232, opened 2021-05-27
Improve display of equal aspect ratio mode in image viewer

astrofrog:
This resolves a long standing issue that when in equal aspect ratio mode the figure used to always be square which did not make an efficient use of the available screen space, and caused issues in tabbed viewer mode, e.g. Jupyter Lab or jdaviz. With the latest change:

Screenshot_2021-05-27_09-53-08

However this isn't ready yet - for now I need to manually call _setup_view_listener after the bqplot image is visible, otherwise the ViewListener does not find the view.

@maartenbreddels - this might be a good starting point for experimenting with possible fixes for ViewListener

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,9 @@
 
 * Use SVG icons instead of PNG for toolbar. [#228]
 
+* Make sure image viewer fills all available space even when using
+  equal aspect ratio. [#231]
+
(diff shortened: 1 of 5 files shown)


maartenbreddels:
https://github.com/glue-viz/bqplot-image-gl/pull/61 should fix this.

PR #238, opened 2021-06-11
FEAT: Deactivate all other tools in other viewers when region selection tool is active in current viewer

### LINKED ISSUE #236: Deactivate all other tools in other viewers when region selection tool is active in current viewer
This is needed for spacetelescope/jdaviz#633 to proceed because currently, when region selection tool is active in multiple viewers at once, the behavior is buggy.

Notes from @astrofrog below:

* Make new intermediate base class for just selections and name it `BqplotSelectionTool` that inherits from `InteractCheckableTool`.
* Overload `activate` method (see snippet below).
* Somehow hook this new tool up back to Jdaviz.

```python
for viewer in self.viewer.session.application.viewers:
    if viewer is not self.viewer:
        viewer.toolbar.active_tool = None

self.viewer.session.edit_subset_mode.edit_subset = None
self.viewer.session.edit_subset_mode.edit_subset = []
```

pllim:
## Description

Fix #236 . This is needed for spacetelescope/jdaviz#633 (Imviz MVP).

Bonus: Make test log readable for a human, and test expectations more consistent between tox and non-tox runs.

### DIFF
--- a/glue_jupyter/bqplot/common/tools.py
+++ b/glue_jupyter/bqplot/common/tools.py
@@ -37,8 +37,8 @@ class BqplotSelectionTool(InteractCheckableTool):
             if viewer is not self.viewer:
                 viewer.toolbar.active_tool = None
 
-        # TODO: Tom asked me to add this, but adding this fails the test.
-        self.viewer.session.edit_subset_mode.edit_subset = []
+        # Jumps back to "create new"
+        self.viewer.session.edit_subset_mode.edit_subset = None
 
         super().activate()
 
--- a/glue_jupyter/bqplot/common/tools.py
+++ b/glue_jupyter/bqplot/common/tools.py
@@ -25,6 +25,24 @@ class InteractCheckableTool(CheckableTool):
         self.viewer._mouse_interact.next = None
 
 
+class BqplotSelectionTool(InteractCheckableTool):
+    """Disables all other tools when activated."""
+
+    def activate(self):
+        with self.viewer._output_widget:
+            self.interact.selected_x = None
+            self.interact.selected_y = None
+
+        for viewer in self.viewer.session.application.viewers:
+            if viewer is not self.viewer:
+                viewer.toolbar.active_tool = None
+
+        # TODO: Tom asked me to add this, but adding this fails the test.
+        # self.viewer.session.edit_subset_mode.edit_subset = []
+
+        super().activate()
+
+
 @viewer_tool
 class BqplotPanZoomMode(InteractCheckableTool):
 
@@ -72,7 +90,7 @@ class BqplotPanZoomYMode(InteractCheckableTool):
 
 
 @viewer_tool
-class BqplotRectangleMode(InteractCheckableTool):
+class BqplotRectangleMode(BqplotSelectionTool):
 
     icon = 'glue_square'
     tool_id = 'bqplot:rectangle'
@@ -126,15 +144,9 @@ class BqplotRectangleMode(InteractCheckableTool):
             if self.finalize_callback is not None:
                 self.finalize_callback()
 
-    def activate(self):
-        with self.viewer._output_widget:
-            self.interact.selected_x = None
-            self.interact.selected_y = None
-        super().activate()
-
 
 @viewer_tool
-class BqplotCircleMode(InteractCheckableTool):
+class BqplotCircleMode(BqplotSelectionTool):
 
     icon = 'glue_circle'
     tool_id = 'bqplot:circle'
@@ -206,12 +218,6 @@ class BqplotCircleMode(InteractCheckableTool):
         if self.interact.selected_x is None or self.interact.selected_y is None:
             if self.finalize_callback is not None:
                 self.finalize_callback()
-
-    def activate(self):
-        with self.viewer._output_widget:
-            self.interact.selected_x = None
-            self.interact.selected_y = None
-        super().activate()
 
 
 @viewer_tool
--- a/glue_jupyter/bqplot/common/tools.py
+++ b/glue_jupyter/bqplot/common/tools.py
@@ -38,7 +38,7 @@ class BqplotSelectionTool(InteractCheckableTool):
                 viewer.toolbar.active_tool = None
 
         # TODO: Tom asked me to add this, but adding this fails the test.
-        # self.viewer.session.edit_subset_mode.edit_subset = []
+        self.viewer.session.edit_subset_mode.edit_subset = []
 
         super().activate()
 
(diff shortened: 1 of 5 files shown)


glue_jupyter/bqplot/common/tools.py
```diff
@@ -25,6 +25,24 @@ def deactivate(self):
...
+        with self.viewer._output_widget:
+            self.interact.selected_x = None
+            self.interact.selected_y = None
+
+        for viewer in self.viewer.session.application.viewers:
+            if viewer is not self.viewer:
+                viewer.toolbar.active_tool = None
+
+        # TODO: Tom asked me to add this, but adding this fails the test.
+        # self.viewer.session.edit_subset_mode.edit_subset = []
```
pllim, line 41: @astrofrog -- What does this line do anyway? The deactivation seems to work fine without it. Enabling this line breaks a test in the CI here.

glue_jupyter/bqplot/common/tools.py
```diff
@@ -207,12 +219,6 @@ def on_selection_change(self, *args):
             if self.finalize_callback is not None:
                 self.finalize_callback()
 
-    def activate(self):
-        with self.viewer._output_widget:
-            self.interact.selected_x = None
-            self.interact.selected_y = None
-        super().activate()
```
astrofrog, line 214: I don't think it makes sense to move these to the base class because ``BqplotXRangeMode`` for example should also inherit from the base class and only has selection in one direction. Having said that I wonder if this code (setting the selected variables to None) should be in the deactivate methods instead of activate.

glue_jupyter/bqplot/common/tools.py
astrofrog: This was the line that should make it change back to 'Create New' - basically edit_subset is the list of subsets currently being modified, and if it's empty it should then become 'Create New'

pllim: Are you expecting the CI to fail with this change? Any idea how to fix?

glue_jupyter/bqplot/common/tools.py
pllim: I have no idea what this thing is doing, so just tell me what to move where and I will do it.

pllim: But I don't understand your comment here. I didn't move it to the global base class, only to `BqplotSelectionTool`, which only rectangle and circle inherit from. How will this affect `BqplotXRangeMode` in any way?

glue_jupyter/bqplot/common/tools.py
```diff
@@ -25,6 +25,24 @@ def deactivate(self):
...
+        with self.viewer._output_widget:
+            self.interact.selected_x = None
+            self.interact.selected_y = None
+
+        for viewer in self.viewer.session.application.viewers:
+            if viewer is not self.viewer:
+                viewer.toolbar.active_tool = None
+
+        # TODO: Tom asked me to add this, but adding this fails the test.
+        self.viewer.session.edit_subset_mode.edit_subset = []
```
pllim, line 41: @astrofrog , I uncommented this line as you requested. CI will fail but I have no idea how to fix it. Please advise. Thanks!

glue_jupyter/bqplot/tests/test_bqplot.py
```diff
@@ -173,20 +173,20 @@ def test_scatter2d_brush(app, dataxyz, dataxz):
     tool2d.interact.selected = [(2.5, 2.5), (3.5, 4.5)]
     tool2d.interact.brushing = False
 
-    assert len(s.layers) == 2
-    assert s.layers[1].layer['x'].tolist() == [3]
-    assert s.layers[1].layer['y'].tolist() == [4]
-    assert s.layers[1].layer['z'].tolist() == [7]
+    assert len(s.layers) == 3
```
pllim, line 176: Okay, I updated the test so it stops failing but I have no idea what these values mean.

pllim:
I installed this branch locally but I don't see any change in the selection behavior in Imviz. 🤷

pllim:
Update: I tried this branch on a different machine with newer GPU and stuff. I still don't see `self.viewer.session.edit_subset_mode.edit_subset = []` doing anything interesting in Imviz. Selection does not go back to "create new" when I open another tiled display.

glue_jupyter/bqplot/common/tools.py
```diff
@@ -25,6 +25,24 @@ def deactivate(self):
...
+        with self.viewer._output_widget:
+            self.interact.selected_x = None
+            self.interact.selected_y = None
+
+        for viewer in self.viewer.session.application.viewers:
+            if viewer is not self.viewer:
+                viewer.toolbar.active_tool = None
+
+        # Jumps back to "create new"
+        self.viewer.session.edit_subset_mode.edit_subset = None
```
pllim, line 41: Okay, `None` does work as you said. It is weird that `[]` is allowed but doesn't do anything. 🤷

glue_jupyter/bqplot/common/tools.py
astrofrog: My point was more that ``BqplotXRangeMode`` is also a selection tool so should also inherit from the base class - however the code I highlighted above is for 2-d selections and won't work for 1-d selectors such as ``BqplotXRangeMode``

astrofrog:
Thanks for working on this! I'm going to push a commit to tweak this a little and will then merge if everything passes.

glue_jupyter/bqplot/common/tools.py
```diff
@@ -19,12 +19,27 @@ def __init__(self, viewer):
         self.viewer = viewer
 
     def activate(self):
+
+        # Disable any active tool in other viewers
+        for viewer in self.viewer.session.application.viewers:
+            if viewer is not self.viewer:
+                viewer.toolbar.active_tool = None
+
```
astrofrog, line 27: I moved this to the base class for all the tools as that then matches what is in the Qt glue package

glue_jupyter/bqplot/common/tools.py
```diff
@@ -19,12 +19,27 @@ def __init__(self, viewer):
...
 
     def deactivate(self):
         self.viewer._mouse_interact.next = None
 
 
+class BqplotSelectionTool(InteractCheckableTool):
+
+    def activate(self):
+        # Jumps back to "create new" if that setting is active
+        if self.viewer.session.application.get_setting('new_subset_on_selection_tool_change'):
```
astrofrog, line 38: I've now made this optional as I think we don't necessarily want that behavior here by default. It can be enabled by passing ``new_subset_on_selection_tool_change=True`` to ``JupyterApplication.__init__`` when used in jdaviz. I've reverted the changes to the tests as the default behavior is now unchanged.

astrofrog (approved):
Looks good now!

pllim:
@astrofrog , not sure what you did, but it does not jump back to "create new" anymore. It does still deactivate other button.

pllim:
Oh, duh... I need that setting... nvm

PR #241, opened 2021-06-29
Add ellipse mode

### LINKED ISSUE #223: Elliptical shape seems allowed in bqplot backend but not really
**Describe the bug**

This implies you support elliptical shape:

https://github.com/glue-viz/glue-jupyter/blob/12f4b2f211f53009dc76b822ba5af7e8db75ebd8/glue_jupyter/bqplot/common/tools.py#L190

But when I use `bqplot:circle` tool over at Jdaviz (Imviz), I can only draw circle.

**Expected behavior**

Not sure what the expectation should be. This is related to spacetelescope/jdaviz#518 , so might need inputs from @eteq et al.

javerbukh:
# Pull Request Template

## Description
This PR attempts to add an EllipticalROI selection tool, which can then be used in jdaviz. It currently uses the `glue_lasso` icon because there is not a `glue_ellipse` icon in glue. 

An alternative to this would be to remove the `pixel_aspect` kwarg in the `BqplotCircleMode` in order to allow ellipses in that mode.

resolves #223

### DIFF
--- a/glue_jupyter/bqplot/common/tools.py
+++ b/glue_jupyter/bqplot/common/tools.py
@@ -215,6 +215,86 @@ class BqplotCircleMode(InteractCheckableTool):
 
 
 @viewer_tool
+class BqplotEllipseMode(InteractCheckableTool):
+
+    icon = 'glue_lasso'
+    tool_id = 'bqplot:ellipse'
+    action_text = 'Elliptical ROI'
+    tool_tip = 'Define an elliptical region of interest'
+
+    def __init__(self, viewer, roi=None, finalize_callback=None, **kwargs):
+
+        super().__init__(viewer, **kwargs)
+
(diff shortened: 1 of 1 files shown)


pllim:
Just curious, does this also fix the shift + drag thingy like with rect/square?

javerbukh:
What is shift + drag supposed to do? For me it locks one axis and lets the other axis move freely to create the shape.

astrofrog (approved):
This works well - I'll push a commit to master with an updated icon. In future we might want to merge this and the circular tool and indeed use shift to toggle between them but this works fine for now.

pllim:
This might have introduced a lag in interactive region creation. When I downgrade to `glue-jupyter` 0.6, I don't see the lag.

pllim:
Nevermind, I think the actual problem is glue-viz/glue#2138 . Also see https://github.com/spacetelescope/jdaviz/issues/723#issuecomment-876594078

PR #244, opened 2021-07-23
Layers dropdown to follow visible Data layer

pllim:
## Description

This is to solve spacetelescope/jdaviz#743 for Jdaviz/Imviz. cc @astrofrog @maartenbreddels 

If this is accepted, we will need a release for Jdaviz.


pllim:
With this patch, I still get something like this when I add markers (i.e., link table data to images). They look confusing but the layer auto selection still works as expected for Imviz. 🤷 

Screenshot 2021-08-10 170038

astrofrog:
Thanks for the PR! The logic isn't quite right yet though - in the following example if I have a subset layer selected and I toggle the visibility of a data layer it will change the selection when it doesn't have to:

Peek 2021-08-12 14-40

I think the logic can actually be simplified. Basically if the current selected layer is visible, leave as is, otherwise select the next visible layer, and if there are none, the next previous visible layer - regardless of whether they are data or subset?

@mariobuikhuizen - on the vue side, I am wondering if we can make things nicer by not actually allowing the user to select an item in the list if it is not visible? Maybe the logic for moving the selection when visibility is toggled could then even be on the vue side too?

pllim:
> I toggle the visibility of a data layer it will change the selection when it doesn't have to

What do you mean? When you hide the second image, the first is shown. When people mess with the options, they want to see things change on what is actually shown. 

While I don't know why the subset stuff is poluting the dropdown, I think the toggle logic is less confusing than it currently is.

If there is a simpler way to do this, I'll be happy too.

astrofrog:
The subset stuff is in the dropdown so that in principle you can change the visual properties of the subsets too. I realised that in jdaviz you may not want this though, so maybe we should have a way to remove subset layers from this dropdown (as an opt-in option that could be enabled in jdaviz).

Layers can have transparency so it doesn't always make sense to select the 'top' visible layer. I think we should only change the selected layer if the *current* one has its visibility turned off (that was the more general point of my animation above). Note that there are viewers (in the general use case) where one might also *only* want to see some subsets and no data hence why I think the logic shouldn't be specific to data vs subset. That issue could be resolved by my previous point that subsets could be hidden in jdaviz if you wanted so there would only be data objects anyway.

pllim:
Well, as far as Imviz is concerned, no one is using the transparency thingy. In fact, they are even confused why there is a Data dropdown box at all. But I understand you have other use cases too, so I'll hold off on this until @astrofrog and @mariobuikhuizen can figure out a path forward. But we do this need for MVP, so a fast resolution would be nice. Thanks!

pllim:
Given #250 has been superseded by #251 , this is not longer merge-able. I converted this to draft so we don't forget about it.

astrofrog:
> Well, as far as Imviz is concerned, no one is using the transparency thingy

Another real use case for the drop down is if you want to overlay contours and images though which is a jdaviz use case (one layer might be the contours and the other the image)

pllim:
Good point. Haven't gotten to fixing contour yet (spacetelescope/jdaviz#530) but it is not MVP, so... 😬

PR #246, opened 2021-08-03
Expand fixed resolution buffer beyond the image boundaries to make panning seamless

astrofrog:
I **think** this will resolve https://github.com/spacetelescope/jdaviz/issues/564

### DIFF
--- a/glue_jupyter/bqplot/image/frb_mark.py
+++ b/glue_jupyter/bqplot/image/frb_mark.py
@@ -2,6 +2,7 @@
 # that implements __call__ with a bounds= argument and returns a fixed
 # resolution buffer (FRB). It is the equivalent of FRBArtist in glue-core.
 
+import math
 import numpy as np
 
 from bqplot import ColorScale
@@ -21,7 +22,7 @@ class FRBImage(ImageGL):
         # FIXME: need to use weakref to avoid circular references
         self.viewer = viewer
 
-        self._external_padding = False
+        self._external_padding = 0
 
         self.scale_image = ColorScale()
         self.scales = {'x': self.viewer.scale_x,
@@ -58,7 +59,7 @@ class FRBImage(ImageGL):
     def external_padding(self, value):
         previous_value = self._external_padding
         self._external_padding = value
-        if value < previous_value:  # no point updating if the value is smaller than before
+        if value > previous_value:  # no point updating if the value is smaller than before
             self.debounced_update()
 
     def update(self, *args, **kwargs):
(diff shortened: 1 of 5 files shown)


pllim:
Does this go with another fix that disable actual pan until mouse button is up? Or this is it?

astrofrog:
I will do a separate PR for the mouse button fix, but you can already try this PR out as it will give you the biggest benefit

glue_jupyter/bqplot/image/frb_mark.py
```diff
@@ -56,6 +56,12 @@ def update(self, *args, **kwargs):
         ymin = self.viewer.figure.axes[1].scale.min
         ymax = self.viewer.figure.axes[1].scale.max
 
+        # Expand beyond the boundary
```
pllim, line 59: What does this mean when the whole image already fits in the view? Does it do unnecessary computation in that case, or is this smart enough?

pllim:
With `glue-jupyter` 0.7:

Screenshot 2021-08-03 132233

With this patch (more performant but display looks grainier):

Screenshot 2021-08-03 131857

pllim:
Maybe the `ny` and `nx` need updating too:

```python
bounds = [(ymin, ymax, ny), (xmin, xmax, nx)]
```

Min/max changed by `ny` and `nx` remained as before, which would explain the degraded resolution.

pllim:
I dunno, maybe something like this? With this, resolution improved and still appeared somewhat performance on drag/zoom, but I did notice a lag in loading (or has it been always there?). 🤷 

```diff
--- a/glue_jupyter/bqplot/image/frb_mark.py
+++ b/glue_jupyter/bqplot/image/frb_mark.py
@@ -65,7 +65,8 @@ class FRBImage(ImageGL):
         if xmin is None or xmax is None or ymin is None or ymax is None:
             return

-        ny, nx = self.shape
+        ny = self.shape[0] + int(dy * 0.5)
+        nx = self.shape[1] + int(dx * 0.5)

         # Set up bounds
         bounds = [(ymin, ymax, ny), (xmin, xmax, nx)]
```

p.s. I also attempted the "only update pan on dragend" but didn't get anywhere. I went as far as setting `PanZoom(..., allow_pan=False` and tried to handle the pan manually but didn't work. I'll leave it to the experts...

glue_jupyter/bqplot/image/frb_mark.py
astrofrog: It is smart enough :)

astrofrog:
@rosteen @pllim - oops good catch about the resolution, should be fixed now. Let me know what the performance is like.

rosteen:
Looks like that resolved the problem I was seeing. The resolution looks good now, and the performance is much better than it was. I still get some white space that takes a moment to fill in occasionally when zoomed in and panning, but it is way better than the constant slow drawing previously.

pllim (approved):
Resolution seems fine now. As @rosteen said, there is still some lag if you pan around or zoom too much real fast, but some improvement is better than no improvement.

pllim:
Hold on... Toolbar is laggy when I re-enable linking in the Imviz example notebook with this patch. Do you also see the same problem?

pllim:
Especially the blinking. Blink is now laggy.

astrofrog:
Yes that could well be - the issue is that the image will now take longer to generate, so blinking could end up being a bit slower. I need to think a little more about how we can improve performance across the board.

astrofrog:
The issue is that this currently slows down the main display of the image, so it would be better to asynchronously load the image outside the field of view *after* the main visible image has been updated.

astrofrog:
I've now made this effectively opt-in for now until we come up with a more elegant solution. To opt in to this, downstream packages should set the ``viewer.state.image_external_padding`` property - for instance, to expand the image by 50% in all directions, this value should be set to 0.5.

pllim:
Thanks, but this isn't really usable for MVP unless you plan to release it in the next 2 days.

astrofrog:
Watch me :laughing:

astrofrog:
See https://github.com/spacetelescope/jdaviz/pull/864 for using this in jdaviz.

PR #247, opened 2021-08-08
Disable resetting of y_min/y_max in layer artist

astrofrog:
I need to figure out if this code is still needed, but for now it seems it definitely isn't for specviz, so this branch can be used with specviz to avoid rescaling all the time.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,5 @@
-0.11 (unreleased)
-=================
+0.10.1 (unreleased)
+===================
 
 * Prevent jumping around of view in profile viewer when creating
   or updating subsets. [#247]
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+0.11 (unreleased)
+=================
+
+* Prevent jumping around of view in profile viewer when creating
+  or updating subsets. [#247]
+
 0.10 (2021-09-14)
 =================
 
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -37,19 +37,22 @@ jobs:
       libraries: {}
       coverage: 'false'
 
-    - linux: py36-test
     - linux: py37-test
+    - linux: py38-test
+    - linux: py39-test
 
-    - macos: py36-test
     - macos: py37-test
+    - macos: py38-test
+    - macos: py39-test
 
-    - windows: py36-test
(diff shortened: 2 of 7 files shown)


pllim:
If it is somehow needed elsewhere, Jdaviz still need an option to turn it off.

pllim:
I can confirm that this also fixes the issue downstream at spacetelescope/jdaviz#760

pllim:
@camipacifici reported that this might have broken the "Home" button, as pressing Home now does not reset Y-axis zoom in Specviz, though it works fine for Imviz.

astrofrog:
I've updated this to a point where I think I'd be happy to merge. I've also fixed the Home button.

At this point it would be helpful if someone could test this out with jdaviz/specviz though?

cc @camipacifici @pllim

rosteen:
I just confirmed that selecting a subset in the profile viewer no longer resets the y-axis zoom, and that the home button now properly resets both axes again in the profile viewer. Looks good!

rosteen (approved):

pllim:
I'll have to double check that the markers no longer reset Imviz zoom/pan, unless someone gets to it first.

javerbukh:
When using mosviz, selecting a new row does not change the Y-axis of the spectrum viewer at all. The scale of the first row persists for all the rest, even if the range is much different.

pllim:
My comment at https://github.com/glue-viz/glue-jupyter/pull/247#issuecomment-895271129 is no longer true. This does *not* prevent pan/zoom reset in Imviz no more when markers are added. Not sure what changed.

But the Imviz issue is not important to fix right now.

astrofrog:
This PR never had any impact on the image viewer so I wouldn't expect it to fix your issue @pllim - but if you open a separate issue I can look into it

pllim:
> This PR never had any impact on the image viewer

It did at one point... 😆 

But don't worry about Imviz for now.

pllim:
But since you asked, I did open an issue at spacetelescope/jdaviz#760

pllim (changes requested):
This actually breaks jdaviz... See spacetelescope/jdaviz#874 . Maybe we need this is a very localized section of Jdaviz only...

astrofrog:
I am working on this in https://github.com/glue-viz/glue/pull/2233

Issue #237, opened 2021-06-11
TST: Handle all the warnings properly


pllim:
As of 2021-06-11, there are 237 warnings in your test suite. The deprecation warnings should be fixed. The other warnings should be handled appropriately. If they are expected, use `pytest.warns`.

dhomeier:
At this point there is probably some pruning of the filters to be done following up on https://github.com/glue-viz/glue-jupyter/pull/248#issuecomment-929329852.
Otherwise remaining warnings right now are the windows-2022 update actions/virtual-environments#4856, no idea if or how this will affect us, and the strange one about xml result files. This looks like there is a
```yaml
    testResultsFiles: '**/test-*.xml'
```
somewhere pulled in, probably from `template: publish.yml@OpenAstronomy`. Do you know anything about that @pllim , and if it should be fixed there or if we should override it in our own `azure-pipelines.yml` (or do we actually want to create some xml files)?

pllim:
Azure template is all @astrofrog and @cadair.

Cadair:
I am confused how the azure templates and the pytest warnings are related?

dhomeier:
https://dev.azure.com/glue-viz/glue-jupyter/_build/results?buildId=1909&view=logs&j=80949b00-1a6f-5532-00f6-7fc3113abd83&t=00ba791c-cdce-595c-8184-e18ca3f742b6&l=11
is apparently running the `PublishTestResults@2` task from `run-tox-env.yml`; question is why is it not finding anything in glue-jupyter, but in glue, astropy etc.?

Cadair:
Oh that error is because it's a docs build so it's not making the coverage xml file the pipelines expect. It's not actually an issue and you can safely ignore it, the other option is to add the `pytest: false` flag to each of these: https://github.com/glue-viz/glue-jupyter/blob/main/azure-pipelines.yml#L56-L58 i.e. this: https://github.com/sunpy/sunpy/blob/main/azure-pipelines.yml#L120

dhomeier:
Thanks. Might need a bit more investigation, since the test, and others like[notebook](https://dev.azure.com/glue-viz/glue-jupyter/_build/results?buildId=1909&view=logs&j=80949b00-1a6f-5532-00f6-7fc3113abd83&t=00ba791c-cdce-595c-8184-e18ca3f742b6&l=11)
[...]

PR #248, opened 2021-08-09
Catch DeprecationWarnings by specific filterwarnings / fixes

dhomeier:
## Description

Addressing #237, this replaces the `-Wignore` pytest option from #238 with more specific `filterwarnings` for the `DeprecationWarning` instances and setting `error` for any that would be missed. All but one actually originate from various upstream modules. So unless I am missing some error in usage of those modules all to be followed up on would be reporting the warnings upstream.
And take care of the two in `glue` itself, of course – `astype(np.float)` is straightforward; replacing `imp` could be a bit trickier.
Getting rid of `np.matrix` in the second commit was also trivial since `Projected3dROI` has already cast `projection_matrix` to `asarray` all along.

### DIFF
--- a/glue_jupyter/ipyvolume/common/tools.py
+++ b/glue_jupyter/ipyvolume/common/tools.py
@@ -22,9 +22,9 @@ class IPyVolumeCheckableTool(CheckableTool):
 
     @property
     def projection_matrix(self):
-        W = np.matrix(self.viewer.figure.matrix_world).reshape((4, 4))     .T
-        P = np.matrix(self.viewer.figure.matrix_projection).reshape((4, 4)).T
-        M = np.dot(P, W)
+        W = np.array(self.viewer.figure.matrix_world).reshape((4, 4))     .T  # noqa: N806
+        P = np.array(self.viewer.figure.matrix_projection).reshape((4, 4)).T  # noqa: N806
+        M = np.dot(P, W)  # noqa: N806
         return M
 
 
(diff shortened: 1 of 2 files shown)


dhomeier:
Well, not sure if it's a realistic endeavour to nail down all potential DeprecationWarnings for all combinations of upstream packages/pyver/OS (with new ones likely popping up as soon as a py38 or py39 test is included)...

Now these runtime warnings popped up, which look actually more serious
```
E   RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility. Expected 80 from C header, got 88 from PyObject
- generated xml file: /home/vsts/work/1/s/.tmp/py37-test/junit/test-results.xml -
```
but why only after adding two more `ignores when half of the tests where completing successfully before???

dhomeier:
Beats me why Windows still ignores the `ignore` for the matplotlib deprecation.
I suggest to keep the `RuntimeWarning`s on binary incompatibility visible, although it's not clear to me if or what can be done about them from our side.

maartenbreddels (approved):

maartenbreddels:
@pllim @astrofrog I wonder if we should merge this, see also #269

pllim (approved):
Definitely an improvement, though follow-up PRs are needed to reduce the number of ignored warnings.

I don't have write access here, so it is all @astrofrog .

maartenbreddels:
With:
 * https://github.com/jupyter-widgets/pythreejs/pull/368
 * https://github.com/bqplot/bqplot/pull/1406
 * https://github.com/glue-viz/glue/pull/2236
 * https://github.com/maartenbreddels/ipyvolume/pull/390
 
We should actually decrease it in time.

astrofrog (approved):
I'll go ahead and merge this, but @dhomeier feel free to open a follow-up PR to remove some of the filters that are no longer needed following the 1.2.4 glue-core release (e.g. the imp one and maybe some of the other ones @maartenbreddels mentioned above).

PR #249, opened 2021-08-10
Avoid duplicate draws in FRBMark

astrofrog:
This does some small optimizations with FRBMark updating to avoid updating too many times when adding a layer

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,8 @@
+0.8 (unreleased)
+================
+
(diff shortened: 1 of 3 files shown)


pllim:
This might have caused a problem with second image not showing up in Layers data drop-down menu as @rosteen reported in https://github.com/spacetelescope/jdaviz/pull/765#issuecomment-896165726

pllim:
I folded a possible fix into #244.

PR #250, opened 2021-08-12
BUG: Fix layers dropdown not showing all loaded data

pllim:
## Description

This is a subset of #244 that only fixes a bug introduced by #249 . Hopefully this is less controversial and can get merged faster. cc @rosteen @astrofrog

### DIFF
--- a/glue_jupyter/widgets/layer_options.py
+++ b/glue_jupyter/widgets/layer_options.py
@@ -72,6 +72,9 @@ class LayerOptionsWidget(v.VuetifyTemplate):
         self.viewer.state.add_callback('layers', _update_layers_from_glue_state)
         _update_layers_from_glue_state()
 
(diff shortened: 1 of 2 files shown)


glue_jupyter/widgets/layer_options.py
```diff
@@ -72,6 +72,9 @@ def _update_layers_from_glue_state(*args):
         self.viewer.state.add_callback('layers', _update_layers_from_glue_state)
         _update_layers_from_glue_state()
```
astrofrog, line 73: I don't fully understand why this isn't doing the trick (the add callback). Let me double check what is going on.

astrofrog:
I think https://github.com/glue-viz/glue-jupyter/pull/251 is the preferred way of fixing this as it gets to why the callback wasn't being triggered at the right time.

pllim:
Superseded by #251

PR #251, opened 2021-08-12
Fix updating of layers list

astrofrog:
Alternative to #250 - we weren't connecting the callback in the right place, and were ending up in a state where the layers state had been updated but not layer artists. This uses private API but in glue-core so this is fine (though we should probably expose this publicly anyway).


astrofrog:
@pllim - can you check if https://github.com/glue-viz/glue-jupyter/pull/251 also fixes the issue with the layer list updating?

pllim (approved):
Yes, it works. Thanks!

pllim:
Turns out this fix is incomplete. It broke the "eye" icon from updating on blink.

PR #268, opened 2021-09-28
feat: truncate text that doesn't fit and show tooltip with full text

mariobuikhuizen:
## Description

Long text in the table is pushing text of other cells out of view:
Screenshot 2021-09-28 at 12 24 48

With this PR the content is truncated and on hover a tooltip appears with the full text:
Screenshot 2021-09-28 at 12 27 01

### DIFF
--- a/glue_jupyter/table/table.vue
+++ b/glue_jupyter/table/table.vue
@@ -9,13 +9,13 @@
       :options.sync="options"
       :items_per_page.sync="items_per_page"
       :server-items-length="total_length"
-      class="elevation-1"
+      class="elevation-1 glue-data-table"
     >
       <template v-slot:header="props">
         <thead>
           <tr>
-            <th style="padding: 0 10px">#</th>
-            <th style="padding: 0 1px" v-if="selection_enabled">
+            <th style="padding: 0 10px; width: 40px">#</th>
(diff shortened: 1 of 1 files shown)


glue_jupyter/table/table.vue
```diff
@@ -63,8 +67,13 @@
 </template>
 
 
-<style scoped>
+<style id="glue_table">
```
maartenbreddels, line 70: why this change? is this for the hot reloading?

maartenbreddels (approved):
Looks good, could you briefly describe the problem from a technical side as well as the solution?

glue_jupyter/table/table.vue
mariobuikhuizen: It's a small fix as `scoped` doesn't work in the ipyvue context and ideally all style tags should have an id to prevent duplicate style tags and leaking old styles during development.

It's not related to this PR, and shouldn't be here.

maartenbreddels: ok, we can leave it in, but good to know it's not related.

mariobuikhuizen:
I found the right way to fix this issue (using `table-layout: fixed;`)

PR #274, opened 2021-12-11
Provide option to reference back to calling app

pllim:
## Description

In Jdaviz, there is no way for the viewer to access the viz app that created it. This PR is needed to remedy that situation.

Needs:

* glue-viz/glue#2256

### DIFF
--- a/glue_jupyter/bqplot/image/viewer.py
+++ b/glue_jupyter/bqplot/image/viewer.py
@@ -32,11 +32,11 @@ class BqplotImageView(BqplotBaseView):
 
     tools = ['bqplot:home', 'bqplot:panzoom', 'bqplot:rectangle', 'bqplot:circle']
 
-    def __init__(self, session):
+    def __init__(self, session, app=None):
 
         super(BqplotImageView, self).__init__(session)
-
         self.shape = None
+        self.app = app
 
         self._composite = CompositeArray()
         self._composite_image = FRBImage(self, self._composite)


astrofrog:
Normally this should be accessible through self.session.application, can you check?

pllim:
I did but that's the JupyterApplication. I need access from viewer to jdaviz.app.Application that's accessible from imviz.app but missing from its viewer instance.

astrofrog:
That seems like a jdaviz-specific thing so not sure it makes sense to implement here, can you not just set .app on the returned instance in jdaviz?

pllim:
I tried that too. It doesn't stick.  I'm not sure why. Maybe overly zealous attribute checks upstream. This is the only way I could make it work without having to do a bunch of stuff to implement backend for viewer specific tray thingies.

astrofrog:
Ok - we have to find a better solution though as other glue-jupyter users won't understand what app is. What if you set session.application.app in jdaviz for instance?

pllim:
That didn't work either. Same problem. It doesn't stick.

astrofrog:
Ohhhh I get it 🤣

maartenbreddels:
I went through the same ideas as @astrofrog , why can't you just set an attribute on session ? like `session.jdaviz = ...` what do you mean by "it doesn't stick" ?

pllim:
I am trying to remember what I did to reach that conclusion. Looks like I tried to do it "properly" by passing `app=...` as part of `**kwargs` that goes into `self._application_handler.new_data_viewer(...)` where `self` is `jdaviz.app.Application`.

But looks like you are suggestion that I simply brute-force it after `viewer` is created. I guess I can give that a try and report back.

pllim:
Nope, doesn't work. See https://github.com/spacetelescope/jdaviz/pull/983/files#r792097008

astrofrog:
Closing in favor of https://github.com/spacetelescope/jdaviz/pull/1051

PR #279, opened 2022-02-09
Only trigger event callbacks on their specified events

Carifio24:
In order to accommodate multiple event callbacks, the mouse interact on the bqplot viewers currently listens for the set of events that is the union of all events specified via calls to `add_event_callback`. Currently, however, when any one of these events is detected, all callbacks are fired, regardless of whether or not it is one of their specified events.

To fix this issue, this PR modifies the viewer's mouse interact listener to check whether the current event is one of the callback's specified events before firing it.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,8 @@
 0.10.2 (unreleased)
 ===================
 
+* Only trigger bqplot viewer callbacks on their specified events. [#279]
+
 * Link opacity of the histogram bars with the layer state in
   the histogram viewer. [#275]
 
--- a/glue_jupyter/bqplot/common/viewer.py
+++ b/glue_jupyter/bqplot/common/viewer.py
@@ -151,7 +151,9 @@ class BqplotBaseView(IPyWidgetView):
 
     def _on_mouse_interaction(self, interaction, data, buffers):
         for callback in self._event_callbacks:
-            callback(data)
+            events = self._events_for_callback.get(callback, [])
+            if data["event"] in events:
+                callback(data)
 
     @debounced(delay_seconds=0.5, method=True)
     def update_glue_scales(self, *ignored):
--- a/glue_jupyter/bqplot/common/viewer.py
+++ b/glue_jupyter/bqplot/common/viewer.py
@@ -1,4 +1,5 @@
 import bqplot
+from functools import partial
 
 from glue.core.subset import roi_to_subset_state
 from glue.core.command import ApplySubsetState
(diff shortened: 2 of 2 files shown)


astrofrog:
@pllim - since this could affect jdaviz, would you be able to check that all the buttons work correctly in imviz with this change?

pllim (changes requested):
Alas, this broke blinking in Imviz and who knows what else.

pllim:
FWIW, there is example value of the variables for my session:

`data`

```python
{'event': 'keydown',
 'pixel': {'x': 777.96875, 'y': 76.609375},
 'domain': {'x': 3312.706444482224, 'y': 1707.172578671637},
 'code': 'KeyB',
 'charCode': 0,
 'key': 'b',
 'keyCode': 66,
 'altKey': False,
 'ctrlKey': False,
 'metaKey': False}
```

`callback`

```python
 functools.partial(<function ImvizImageView.on_mouse_or_key_event at ...,
                   <jdaviz.configs.imviz.plugins.viewers.ImvizImageView object at ...>)
```

Carifio24:
So the reason this broke jdaviz is because I had forgotten to account for the fact that the items yielded by `CallbackContainer.__iter__` aren't the same as what are inserted via `append`. Hence the lookup in the events dictionary was always giving back an empty list.

I think my most recent commit should fix this. The `CallbackContainer` yields a new `partial` on every run through `__iter__`, so to get around this, the key that's stored in the events dictionary (when the callback is a method or a partial) is a tuple `(callback.__func__, (callback.__self__))`, which can easily be reproduced from a partial function `p` as `(p.func, p.args)`. This is then the key that's used in `_on_mouse_interaction`. @astrofrog does this seem like a reasonable way to handle this?

pllim:
Seems to work for me now in limited testing. Thanks!

astrofrog (approved):
Thanks!

pllim:
Oh, no. I think this broke Jdaviz after all. I should have ran the whole CI, my bad.

https://github.com/spacetelescope/jdaviz/runs/5325959132?check_suite_focus=true

pllim:
Is it possible to revert this and do a 0.11.1 release until we can figure it out? 🙏

astrofrog:
Can you provide more details of the failure? I can't do a bug fix immediately but will see if I can do it tomorrow. However I want to understand first whether it is actually an issue here or an issue in how jdaviz was using this functionality.

astrofrog:
Ah I see it is in https://github.com/spacetelescope/jdaviz/issues/1117 - @Carifio24 do you think this would be an easy fix?

pllim:
As you can see in spacetelescope/jdaviz#1117 ,  Specviz has a cube slicing plugin where you can click on the 1D spectrum and the cube image viewer would show the associated 2D image slice. That part of the code seems to be broken with this patch. It is unclear what is happening in the traceback without more digging.

Carifio24:
Based on that traceback I think I see the problem (looks like it's here, and not with jdaviz), and I also think it should be a quick fix. I'll submit a fix tonight if I can, otherwise first thing tomorrow.

pllim:
Thank you! 🙏

PR #284, opened 2022-02-15
Throttle set color calls in color pallette

kecnry:
## Description

This PR implements throttling in JS to avoid having to wait for the color to catch up when dragging in the color palette by throttling emitting the events to python to once every 100ms.  This throttling time could be fine-tuned (ideally want to be the cost of a single update) or made dynamic based on the number of layers/datapoints.

### DIFF
--- a/glue_jupyter/widgets/layeroptions.vue
+++ b/glue_jupyter/widgets/layeroptions.vue
@@ -23,7 +23,7 @@
                                     <v-icon>mdi-close</v-icon>
                                 </v-btn>
                                 <v-color-picker :value="data.item.color"
-                                            @update:color="set_color({index: data.item.index, color: $event.hex})"></v-color-picker>
+                                            @update:color="throttledSetColor({index: data.item.index, color: $event.hex})"></v-color-picker>
                             </div>
                         </v-menu>
                         <v-btn icon @click.stop="toggle_visible(data.item.index)">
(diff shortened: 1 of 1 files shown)


pllim:
Can't say I notice much difference or am I not supposed to?

kecnry:
Are you seeing it equally laggy before and after or just aren't reproducing the "catch up" lagginess at all with your case?  I think in most cases the lagginess isn't really noticeable and so this won't make much of a difference, but as the time to update the color increases, then this will start to help (until the update time takes longer than the throttle time in which case it'll start showing up again).

pllim:
Not sure. I don't use the color thingy. Maybe we should wait for other devs to also have a look.

pllim:
I think it is slightly smoother here but I can't tell if I am imagining it or not. 😆 

Also, it works in Lab dark mode! 😉

astrofrog:
I'll try and test this shortly, thanks!

pllim:
Is this a duplicate of #313 to fix #312?

mariobuikhuizen:
> jupyter-widget will not send an update to a trait (from the frontend to the backend) when an update is still in flight. If we use a custom message, we will flood the kernel with messages (that it probably cannot catch up with).

I've tried with using a trait, but that still causes an update loop

maartenbreddels:
> I've tried with using a trait, but that still causes an update loop

For the record, we dove into this, see https://github.com/glue-viz/glue-jupyter/pull/313#issuecomment-1231782625  and the reason the trait caused a loop as a bug in ipywidgets.

in #313 we will continue to fix, so we can close this issue.

Thanks for the work though @kecnry , it helped us understand what was going on!

Issue #282, opened 2022-02-15
A way for developer to see what messages are fired from Glue


pllim:
As @astrofrog suggested today. Motivation: Developers cannot see which message are being fired by events in Jdaviz during debugging.

PR #286, opened 2022-02-17
Add note to developer docs about logger

astrofrog:
Fixes https://github.com/glue-viz/glue-jupyter/issues/282

### DIFF
--- a/docs/developer_notes.rst
+++ b/docs/developer_notes.rst
@@ -96,3 +96,17 @@ components of glue:
 * :ref:`glue:data_tutorial` - for the data collection, data, and subsets
 * :ref:`glue:linking-framework` - for the data linking framework
 * :ref:`glue:communication` - for the event framework
+
+Inspecting glue messages being broadcast
+----------------------------------------
+
+When performing certain actions (e.g. updating a subset), glue will emit
+messages which can be listened for (see ... for more details). It can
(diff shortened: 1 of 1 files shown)


docs/developer_notes.rst
```diff
@@ -96,3 +96,17 @@ components of glue:
...
+
+When performing certain actions (e.g. updating a subset), glue will emit
+messages which can be listened for (see ... for more details). It can
+sometimes be helpful to see all the messages being broadcast - in a glue-jupyter
+this can be done by calling::
+
+    from glue.logger import logger
+    logger.setLevel('INFO')
+
+After this call, you will see when clients subscribe to the hub for messages
```
pllim, line 111: When this is done from Jdaviz in a notebook, should we be looking at the notebook cell or terminal?

pllim:
It is in the notebook cell output but I am not sure how useful it is for me as a downstream developer. I see *everything*, not just the messages I am interested in. When you have a bunch of messages that won't go away in the same cell output as the image display, it is very hard to test things interactively. 🤷 

Screenshot 2022-02-18 152938

Also for whatever reason, the log message does not appear together with the viewer as shown above when you use Lab instead of notebook. It still shows for other cells though. 🤷

astrofrog:
I'm going to go ahead and merge this for now and then will work on a better solution to show messages.

PR #289, opened 2022-02-22
Fix eye icon toggling for layers

astrofrog:
Fixed as part of the second jdaviz hack day


pllim:
So I found an old comment: https://github.com/spacetelescope/jdaviz/issues/897#issuecomment-924395409

Turns out you are adding back the line removed in #251.

PR #293, opened 2022-02-22
BUG: Fix out of bounds when viewer shape is zero

pllim:
## Description

Sometimes, `self.shape` can have zeroes and causes out of bounds error.

Fix spacetelescope/jdaviz#1102

Pairs well with #270 for the other common error when dealing with subset deletions, and maybe red wine.


astrofrog (approved):

astrofrog:
No I'm not sure, we should investigate but the fix here is good enough for now

PR #295, opened 2022-02-24
Revert "Only trigger event callbacks on their specified events"

pllim:
Reverts glue-viz/glue-jupyter#279

Broke Jdaviz.


pllim:
CI passed at spacetelescope/jdaviz#1118 , which uses this PR, which proves #279 is indeed the cause.

PR #299, opened 2022-03-17
fix: prevent long content from pushing out of view

mariobuikhuizen:
# Pull Request Template

## Description

Prevent long content from pushing out of view. See https://github.com/spacetelescope/jdaviz/issues/1157.

### DIFF
--- a/glue_jupyter/common/state_widgets/viewer_image.vue
+++ b/glue_jupyter/common/state_widgets/viewer_image.vue
@@ -1,21 +1,21 @@
 <template>
-    <div>
+    <div class="glue-viewer-image">
         <div>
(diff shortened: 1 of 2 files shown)


pllim:
I am not sure how but I think this broke some stuff in Jdaviz. For example: The slider indicator does not work anymore (though the slicing works by updating the slice as you click on Cubeviz spectrum viewer but the indicator orange line does not move), and some eye icons (not all of them) stopped working.

pllim:
Nvm, I think that weird behavior also exists without this patch.

pllim (approved):
@astrofrog , we need this. Thanks!

PR #300, opened 2022-03-29
Migrate CI tests and publisher from Azure Pipelines to GitHub Actions

dhomeier:
## Description
Transition from Azure pipelines using OpenAstronomy workflows with otherwise mostly unchanged envs for tests, wheel and sdist build; adding the experimental `macos-12-aarch64` target for `py310`.


dhomeier:
@astrofrog since even the more problematic macOS runs did succeed at one point in https://github.com/glue-viz/glue-jupyter/actions/runs/2058744911 all I can think of right now is waiting some time until GH Actions may have come back to their senses.
One change here that might be relevant for a release is however that I've changed the minimum versions in setup.cfg since the new jinja2 3.1.1 breaks any nbconvert < 6.4.5. Alternatively we'd have to pin jinja2 < 3.1, but the newer versions did pass earlier (and do on [Azure](https://dev.azure.com/glue-viz/glue-jupyter/_build/results?buildId=2031&view=logs&j=0a22f5d8-9f3e-5735-4672-5a55fb1a2abf&t=7886e472-715d-500e-d95f-9ccab8aa9bc3)), and jupyter-client is already at 7.2.
Perhaps put that into a separate PR?

astrofrog:
@dhomeier - I've included the nbconvert fix in https://github.com/glue-viz/glue-jupyter/pull/301 - thanks!

dhomeier:
and I have set up the wheel build as `publish_pure_python` here; don't know if what
https://github.com/glue-viz/glue-jupyter/runs/5738236254?check_suite_focus=true
has produce is the right thing...

.github/workflows/ci_workflows.yml
```diff
@@ -0,0 +1,64 @@
...
+  publish:
+    uses: OpenAstronomy/github-actions-workflows/.github/workflows/publish_pure_python.yml@v1
+    with:
+      test_extras: test
+      test_command: pytest --pyargs glue_jupyter
+    secrets:
+      pypi_token: ${{ secrets.pypi_token }}
+
+  allowed_failures:
+    # some (all?!?) macOS targets are failing on GH actions without any output or explanation
```
astrofrog, line 58: I think these are working now, can you add this job back up to the main ones?

astrofrog (changes requested):

dhomeier:
Something came out of the rebase backwards – git told me there were newer commits here; hope that's not creating too much confusion here.

astrofrog:
@dhomeier - I wonder if you rebased on master instead of main? I've rebased again and it has got rid of the extra commits.

dhomeier:
Thanks! I don't think it was master, but I may have rebased on the branch again before just pushing --force.

astrofrog (approved):
Thanks!

PR #301, opened 2022-03-29
Fix how we handle the default ROIClickAndDrag to use the new interact event infrastructure

astrofrog:
This was previously overwriting the top level MouseInteract rather than using 'next' to set the temporary ROI interact tools. This also switches to listen to only dragstart events using the new ``add_event_callback`` machinery.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+0.11.3 (unreleased)
+===================
+
+* Fixed a bug that caused clicking and dragging of existing regions to not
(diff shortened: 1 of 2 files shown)


rosteen (approved):
I confirmed that this fixes the bug I was seeing in Jdaviz. Looks good to me (well, other than the test failures of course).

javerbukh (approved):
Fixes the bug and gives us a cool new (working) feature, thanks!

pllim:
:shipit:

PR #302, opened 2022-03-29
feat: make compatible with future versions of bqplot

maartenbreddels:
A future version of bqplot (0.13.0) will split off the webgl part
into a bqplot-gl package.

This will make it work with that future version, I'm just not sure how we need to optionally depend on bqplot-gl

### DIFF
--- a/glue_jupyter/bqplot/compatibility.py
+++ b/glue_jupyter/bqplot/compatibility.py
@@ -1,0 +1,13 @@
+from bqplot_image_gl import ImageGL
+import bqplot
+
+ScatterGL = None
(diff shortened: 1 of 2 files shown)


astrofrog:
What about just adding bqplot-gl as a required dependency? It will only get used once bqplot itself no longer has ScatterGL which will only happen once there is a stable release of bqplot-gl I imagine?

maartenbreddels:
bqplot-gl will have a dependency on bqplot 0.13. So we can require that, but at least this will make the current version compatible with future versions, of bqplot-gl is installed along with it.

astrofrog:
I am going to rebase this to force the CI to re-run

PR #303, opened 2022-03-31
fix: do not reformat float numbers in text fields

maartenbreddels:
A roundtrip can cause 1e2 to directly reformat to 100,
instead, we now try to keep the original formatting in the frontend.

Fixes https://github.com/spacetelescope/jdaviz/issues/1188

cc @PatrickOgle


astrofrog (approved):
Looks great!

PR #304, opened 2022-04-04
expose axes limits in profile viewer options

kecnry:
# Pull Request Template

## Description

This pull request adds axes limits to the viewer options for a profile viewer.  In theory this could be extended to other viewers, although handling aspect ratio within image viewers will require additional logic (currently the aspect ratio is enforced at every keypress, which overwrites the value typed by the user, making the axes limit editing effectively useless while equal aspect ratio is enabled).


mariobuikhuizen:
The solution for the problem with margins inside the jdaviz sidebar is to use `<v-row no-gutters>`

mariobuikhuizen:
And add some margin between the glue-float-fields with `class="mr-1"` and `class="ml-1"`

astrofrog (changes requested):
This is a great start, thanks! It would indeed be nice for the limits to be updated only once the user has finished entering values, otherwise typing e.g. 1000 means the limits will change four times (1, then 10, then 100, then 1000). Maybe @mariobuikhuizen and @maartenbreddels have ideas how to deal with this? I think this would solve the aspect ratio issue you mention for the image viewer.

astrofrog:
@kecnry I've pushed some commits here to add the 'Apply limits' button. Do you think this seems ok?

glue_jupyter/widgets/axes_limits.py
```diff
@@ -0,0 +1,38 @@
...
+        self._cache = (self.x_min, self.x_max, self.y_min, self.y_max)
+        with delay_callback(self.glue_state, 'x_min', 'x_max', 'y_min', 'y_max'):
+            if self.x_min is not None:
+                self.glue_state.x_min = self.x_min
+            if self.x_max is not None:
+                self.glue_state.x_max = self.x_max
+            if self.y_min is not None:
+                self.glue_state.y_min = self.y_min
+            if self.y_max is not None:
+                self.glue_state.y_max = self.y_max
```
kecnry, line 38: when used in an image viewer (now enabled), how is a change in _all_ limits, with equal aspect enabled, handled with this delayed callback?  From testing, it seems that the user-set x-limits are preserved and the y-limits are somehow adjusted to enforce the equal aspect?

I'm honestly not sure what the expected behavior should be in this case... maybe the smallest bounding box in x and y that contains the user-provided values?

Issue #134, opened 2019-09-30
Create glue-jupyter viewer registry


nmearl:
The viewer registry is necessary for interactions with gui-based applications created using glue-jupyter. Currently, glue-jupyter does not provide the importable registry.

astrofrog:
We should move the registry defined in jdaviz:

https://github.com/spacetelescope/jdaviz/blob/c020afdb2dcba5ea97a80c3140e55f4ff8f3224d/jdaviz/core/registries.py#L51

into glue-jupyter! :smile:

PR #307, opened 2022-04-08
Add registry for adding viewers from plugins

jfoster17:
# Pull Request Template

## Description

This PR adds a registry of viewer types (analogous to `qt_client` in QTglue). It allows an external plug-in to define a new viewer that is then available to the glue_jupyter application `new_data_viewer` function. The new viewer class is registered via a decorator which allows the developer to specify a nice string name for ease of creating a new viewer.

`app = gj.jglue()`
`s = app.new_data_viewer('externalviewer')`

This PR addresses #134 and mostly uses the jdaviz code referenced therein for the `ViewerRegistry` logic.

### DIFF
--- a/glue_jupyter/app.py
+++ b/glue_jupyter/app.py
@@ -151,6 +151,18 @@ class JupyterApplication(Application):
         return viewers
 
     def new_data_viewer(self, *args, **kwargs):
+        """
+        Create a new data viewer
+
+        This function can be called directly with the name of a viewer
+        as the first parameter for any viewer types registered in the
+        viewer_registry.
+
+        s = app.new_data_viewer('table')
+
+        This is the preferred way to call viewers defined in external plugins.
+        Built-in viewers can also be called using the custom functions
+        """
         show = kwargs.pop('show', True)
         viewer_name = args[0]
         if isinstance(viewer_name, str):
--- a/glue_jupyter/app.py
+++ b/glue_jupyter/app.py
@@ -17,6 +17,8 @@ from glue_jupyter.utils import _update_not_none, validate_data_argument
 from glue_jupyter.widgets.subset_select_vuetify import SubsetSelect
 from glue_jupyter.widgets.subset_mode_vuetify import SelectionModeMenu
 
+from glue_jupyter.registries import viewer_registry
+
 __all__ = ['JupyterApplication']
 
 # TODO: move this to glue-core so that the subset mode can be set ot a string
@@ -150,6 +152,12 @@ class JupyterApplication(Application):
 
     def new_data_viewer(self, *args, **kwargs):
         show = kwargs.pop('show', True)
+        if isinstance(args[0], str):
+            try:
+                viewer_cls = viewer_registry.members[args[0]]['cls']
+            except KeyError:
+                raise ValueError("No registered viewer found with name {0}".format(args[0]))
+            args = (viewer_cls, )
         viewer = super().new_data_viewer(*args, **kwargs)
         self._viewer_refs.append(weakref.ref(viewer))
         if show:
--- a/glue_jupyter/app.py
+++ b/glue_jupyter/app.py
@@ -156,12 +156,18 @@ class JupyterApplication(Application):
 
         This function can be called directly with the name of a viewer
         as the first parameter for any viewer types registered in the
-        viewer_registry.
+        viewer_registry. Thus if a plug-in defines a viewer class as
 
-        s = app.new_data_viewer('table')
+        from glue_jupyter.registries import viewer_registry
+        @viewer_registry("pluginviewer")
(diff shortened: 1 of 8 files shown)


jfoster17:
The ViewerRegistry I copied over from jdaviz also defines a label for the viewer which currently is not used by glue_jupyter, but seems like it might be useful for future glue-jupyter-lab development. Should I remove it?

glue_jupyter/app.py
```diff
@@ -150,6 +152,12 @@ def viewers(self):
 
     def new_data_viewer(self, *args, **kwargs):
         show = kwargs.pop('show', True)
+        if isinstance(args[0], str):
+            try:
+                viewer_cls = viewer_registry.members[args[0]]['cls']
+            except KeyError:
```
astrofrog, line 158: Technically this might also fail if ``['cls']`` is not found - maybe instead of try...except you could simply check ``if args[0] in viewer_registry.members``?

glue_jupyter/registries.py
```diff
@@ -0,0 +1,39 @@
+from glue.config import DictRegistry
+
+__all__ = ['viewer_registry', 'ViewerRegistry']
+
+
+class ViewerRegistry(DictRegistry):
+    """
+    Registry containing references to custom viewers.
```
astrofrog, line 8: I wonder if this doesn't need to say 'custom' viewers as maybe we could actually add the regular viewers to it too so that they could also be created with the same API if desired?

astrofrog (changes requested):
Looks good overall, just a couple of comments. Maybe just remove the label stuff for now as you suggest?

glue_jupyter/registries.py
jfoster17: That's a good idea, but I have some questions/comments about the preferred way to do this.

1. Currently glupyter does not process plugin entry points from glue-jupyter (just from glue/glue-core). The built-in viewers are just imported in the functions in `app.py`. So in commit https://github.com/glue-viz/glue-jupyter/pull/307/commits/d214fe72cbdcaa40f46dae3d7b5a2aa794533cbe I have to import the viewers into the test functions before I can call them using the new (decorated) interface. The regular glue way is to add the viewers to the registry when their `setup` method runs from processing entry_points and all the default glue viewers have entries here. 
2. Since we have both bqplot and matplotlib viewers, the names for viewers are either ugly or default. That is, it probably makes sense to keep the bqplot viewers are default, so the name `scatter` means the bqplot scatter and if you wanted the matplotlib scatter we would call that `matplotlib-scatter` or something unintuitive like that.

glue_jupyter/app.py
jfoster17: Okay, I've made this a bit more robust (and added tests) in https://github.com/glue-viz/glue-jupyter/pull/307/commits/2300589b01829e58f2d1c64225835cd921cbe21b

glue_jupyter/registries.py
jfoster17: So, in summary -- this PR currently works fine for my current desire to add a new viewer to glue-jupyter via a separate plugin (this plugin calls glue.entry_points to add the viewer to the registry). This PR does NOT YET add all the existing viewers to the registry to enable them to be called via:

`app.new_data_viewer('table')`

although it could. My main hesitation (apart from the naming question I reference above) is that this `new_data_viewer` interface does not support any configuration options at creation time (e.g. `x = some_attribute`). I could extend it to do so -- we basically try to create state objects from all kwargs except `show` and `data` -- but that is a more significant change to the API if any other code is already calling `new_data_viewer`. This new method for creating viewers will always be less desirable from a notebook because you don't have the nice custom docstring that the dedicated viewer creation functions (e.g. `histogram_1d`) have for helping the user know what parameters are available at viewer creation time so perhaps it's not worth the effort?

setup.cfg
```diff
@@ -43,6 +43,11 @@ docs =
     nbsphinx
     sphinx-rtd-theme
 
+[options.entry_points]
+glue.plugins =
+    table_viewer = glue_jupyter.table:setup
```
astrofrog, line 48: This doesn't exist, right? (the ``setup`` function). I agree with what you said in your other comments about not doing this yet for build-in viewers, so maybe remove for now.

astrofrog (changes requested):
This is great! Aside from my comment below, could you:

* Mark this PR as 'ready for review' (so that it's not a draft)
* Add a short section in https://glue-jupyter.readthedocs.io/en/latest/developer_notes.html to mention the existence of this registry and giving an example

Once this is done we can go ahead and merge it. Thanks!

jfoster17:
Okay, I think we're all set @astrofrog !

astrofrog (approved):
Perfect, thanks!

PR #309, opened 2022-04-19
as_steps support in profile viewer

kecnry:
## Description

This accompanies https://github.com/glue-viz/glue/pull/2292 to expose the `as_steps` switch in the layer options widget, and handling toggling `as_steps` in the bqplot layer artist for the profile viewer.  The actual binning logic is adopted from https://github.com/astropy/specutils/blob/main/specutils/spectra/spectral_axis.py#L42.

### DIFF
--- a/glue_jupyter/common/state_widgets/layer_profile.vue
+++ b/glue_jupyter/common/state_widgets/layer_profile.vue
@@ -7,7 +7,7 @@
             <v-select label="attribute" :items="attribute_items" v-model="attribute_selected" hide-details />
         </div>
         <div>
-            <v-switch label="Plot as steps/histogram" v-model="as_steps" />
+            <v-switch label="Plot as steps" v-model="as_steps" />
         </div>
     </div>
 </template>
--- a/glue_jupyter/common/state_widgets/layer_profile.vue
+++ b/glue_jupyter/common/state_widgets/layer_profile.vue
@@ -6,5 +6,8 @@
         <div>
             <v-select label="attribute" :items="attribute_items" v-model="attribute_selected" hide-details />
         </div>
+        <div>
+            <v-switch label="Plot as steps/histogram" v-model="as_steps" />
+        </div>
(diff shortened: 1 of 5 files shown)


glue_jupyter/common/state_widgets/layer_profile.vue
```diff
@@ -6,5 +6,8 @@
         <div>
             <v-select label="attribute" :items="attribute_items" v-model="attribute_selected" hide-details />
         </div>
+        <div>
+            <v-switch label="Plot as steps/histogram" v-model="as_steps" />
```
astrofrog, line 10: I would avoid using the terminology 'histogram' here, maybe just use 'steps'

astrofrog (changes requested):
As mentioned in https://github.com/glue-viz/glue/pull/2292#issuecomment-1102814301, I think we should instead be implementing the logic of converting the profile into a stepped profile in the bqplot layer artist so that it would be analogous to the fact that this conversion is done in the layer artist for Matplotlib (in fact it is done inside Matplotlib because of the drawstype='steps' option).

astrofrog (changes requested):
Looks good and works great! Can you add a changelog entry? Also we will need to wait to merge this until a new version of glue-core is released, will try and do shortly once we merge the other PR.

kecnry:
Just noticed that this seems to break when adding a subset (within the bqplot steps logic)... we'll need to fix that before merging/releasing.

astrofrog (changes requested):
Could you check that this works with glue-core 1.3.0 and if so bump the minimum glue-core version in setup.cfg? Once that is done I think we can merge this.

kecnry:
Sorry this fell off the radar for a little while.  I rebased, confirmed this works with glue 1.3+, and set that as a requirement in setup.cfg.

astrofrog (approved):
Looks great, thanks!

PR #311, opened 2022-05-03
Initial support for unit conversion in profile viewer

astrofrog:
This is a proof of concept of unit conversion in the profile viewer. It is a companion to the following PR in glue-core:

* https://github.com/glue-viz/glue/pull/2296

(that PR is needed for the changes here to work)

For now, this adds dropdowns for the units for the x and y axis, though note that I have not yet implemented the x axis unit conversion. The combo for the y axis currently starts off empty which means 'native units', i.e. no unit conversion, and it is then possible to select a unit. This is demonstrated in the following notebook:

https://gist.github.com/astrofrog/715239c5fb7b015019fdb414ae2b609e

Obviously we should think about what the default should be, and I think we should at least indicate in the 'no conversion' case something like 'native/data units' or similar. My plan going forward is the following:

* [x] Implement unit conversion for the x-axis too
* [x] Make sure x range selections work properly
* [x] Try and make the ability to specify alternative units pluggable so that e.g. jdaviz can opt in to using e.g. the spectral equivalency for the x-axis.
* [x] Add tests/docs/etc

Anyway, I'm putting this up now in case there is any early feedback/ideas!

### DIFF
--- a/notebooks/Generic/demo_units_custom.ipynb
+++ b/notebooks/Generic/demo_units_custom.ipynb
@@ -1,0 +1,334 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This notebook demonstrates unit conversion in the profile viewer"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "from astropy.wcs import WCS\n",
+    "import glue_jupyter as gj\n",
+    "from glue.core import Data\n",
+    "from glue.plugins.wcs_autolinking.wcs_autolinking import WCSLink"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Set up two datasets with x-axes that use different spectral coordinates (which get auto-converted because of the WCSLink) and y-axes that use different temperature units (just to show how to add custom equivalencies)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "wcs1 = WCS(naxis=1)\n",
+    "wcs1.wcs.ctype = ['WAVE']\n",
+    "wcs1.wcs.crval = [10]\n",
+    "wcs1.wcs.cdelt = [3]\n",
+    "wcs1.wcs.crpix = [1]\n",
+    "wcs1.wcs.cunit = ['cm']\n",
+    "\n",
+    "d1 = Data(temp1=np.random.random(10))\n",
+    "d1.get_component('temp1').units = 'deg_C'\n",
+    "d1.coords = wcs1"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "wcs1 = WCS(naxis=1)\n",
+    "wcs1.wcs.ctype = ['FREQ']\n",
+    "wcs1.wcs.crval = [1000]\n",
+    "wcs1.wcs.cdelt = [1000]\n",
+    "wcs1.wcs.crpix = [1]\n",
+    "wcs1.wcs.cunit = ['MHz']\n",
+    "\n",
+    "d2 = Data(temp2=np.random.random(10) + 273)\n",
+    "d2.get_component('temp2').units = 'K'\n",
+    "d2.coords = wcs1"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Set up a custom unit conversion class with equivalencies:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from astropy import units as u\n",
+    "from glue.core.units import unit_converter\n",
+    "\n",
+    "@unit_converter('custom-astropy')\n",
+    "class UnitConverterWithTemperature:\n",
+    "\n",
+    "    def equivalent_units(self, data, cid, units):\n",
+    "        equivalencies = u.temperature() if 'temp' in cid.label.lower() else None\n",
+    "        return map(u.Unit(units).find_equivalent_units(include_prefix_units=True, equivalencies=equivalencies), str)\n",
+    "\n",
+    "    def to_unit(self, data, cid, values, original_units, target_units):\n",
+    "        equivalencies = u.temperature() if 'temp' in cid.label.lower() else None\n",
+    "        return (values * u.Unit(original_units)).to_value(target_units, equivalencies=equivalencies)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Use the converter class defined above"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from glue.config import settings\n",
+    "settings.UNIT_CONVERTER = 'custom-astropy'"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/home/tom/Code/glue/glue/external/echo/__init__.py:3: UserWarning: glue.external.echo is deprecated, import from echo directly instead\n",
+      "  warnings.warn('glue.external.echo is deprecated, import from echo directly instead')\n",
+      "/home/tom/Code/glue/glue/viewers/common/qt/tool.py:5: UserWarning: glue.viewers.common.qt.tool is deprecated, use glue.viewers.common.tool instead\n",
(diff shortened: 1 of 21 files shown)


astrofrog:
I've now added x-axis unit conversion though I still need to fix the selection - no changes in this PR but the x-axis conversion was added in the glue-core PR. I've updated the gist: https://gist.github.com/astrofrog/715239c5fb7b015019fdb414ae2b609e

astrofrog:
I'll work on fixing the ROI selection when I get back to this, it should be just a one-line fix.

notebooks/Generic/demo_units_custom.ipynb
```diff
@@ -0,0 +1,334 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 6,
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "/home/tom/Code/glue/glue/external/echo/__init__.py:3: UserWarning: glue.external.echo is deprecated, import from echo directly instead\n",
```
pllim, line 120: You probably want to clear the outputs if you plan to keep this notebook at merge time.

pllim (commented):
Pretty cool at a glance! Do you plan to hook this up to Jdaviz? I think it is clearer to see if this is enough or not when we can test with some existing Jdaviz data and what-nots.

Screenshot 2022-05-12 151054

Unclear to me why X-axis only has "Wave" as label without unit and Y-axis label is completely missing. I don't see you set either of them explicitly here. So it is hard to say whether the display axis is actually displaying the correct stuff or not at a glance. (Though upon looking at the code, I think they are correct.)

notebooks/Generic/demo_units_custom.ipynb
```diff
@@ -0,0 +1,334 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 5,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from glue.config import settings\n",
+    "settings.UNIT_CONVERTER = 'custom-astropy'"
```
kecnry, line 108: would there ever be a use-case where you'd want to use different converters for different app/viewer instances simultaneously?  If so, being able to pass the converter (or its name) to a lower-level API to override the package-level setting might be useful.

pllim: Won't that be super confusing though in terms of reproducible science, unless there is a way to track which converter was used to convert what?

kecnry: I don't think so, it would still either be explicitly passed or defined by the viewer itself.  I'm thinking of cubeviz/mosviz where we have a spectral viewer and image viewer in the same glue app and we probably only want one of those 4 axes to use spectral equivalancies.

pllim: As far as `astropy` is concerned, it doesn't hurt to enable equivalencies even if you never uses it, so in theory you can enable the same set of equivalencies for all the viz. The more complicated part is when you have to pass inputs to some of them but even then maybe we can make use of helper functions like `synphot`, maybe.

```python
>>> from astropy import units as u
>>> x = 1 * u.m
>>> x.to(u.cm)
<Quantity 100. cm>
>>> x.to(u.cm, u.spectral())
<Quantity 100. cm>
>>> x.to(u.cm, u.spectral_density(x))
<Quantity 100. cm>
```

kecnry: But you probably don't want to expose frequency units as choices on a non-spectral length axes (and might even want to raise an error if the user requests display units in frequency).

pllim: I cannot think of an example in Imviz where that concern would be an issue. @kecnry , what is the use case that you worry about?

astrofrog: The API of the converter gives access to the component ID during the conversion so you can always decide what equivalencies to include based on that? (See simplistic example in the docs in the glue-core PR)

pllim: This one?

```python
    from astropy import units as u
    from glue.core.units import unit_converter

    @unit_converter('example-2')
    class ExampleUnitConverter:

        def equivalent_units(self, data, cid, units):
            equivalencies = u.temperature() if 'temp' in cid.label.lower() else None
            return map(u.Unit(units).find_equivalent_units(include_prefix_units=True, equivalencies=equivalencies), str)

        def to_unit(self, data, cid, values, original_units, target_units):
            equivalencies = u.temperature() if 'temp' in cid.label.lower() else None
            return (values * u.Unit(original_units)).to_value(target_units, equivalencies=equivalencies)
```

astrofrog: Yes - in practice you wouldn't really want to use just the name of the label, but you could imagine using e.g. info from the header, or the WCS, to make decisions about what equivalencies to use.

astrofrog:
@pllim @kecnry - aside from the comments above, do you think this would be suitable for jdaviz? If so, I will try and wrap it all up and merge the glue-core and this PR.

pllim:
I wasn't involved in the Unit Conversion plugin, so I am going to defer to @kecnry , @rosteen , @javerbukh , or @duytnguyendtn .

kecnry:
Yes, I think so, but haven't actually fleshed out how exactly it would be implemented in jdaviz yet.  If you'd prefer we do that before you merge in case that would motivate any changes, let us know and we can try to get to that soon.

astrofrog:
I guess I want to make sure this approach is at least compatible with what jdaviz wants as if I end up having to implement something quite different it wouldn't be great to have multiple different ways of dealing with units. But if you think it is the way we want to go I am happy to wrap up and merge.

pllim:
Should Jdaviz do a proof of concept against this PR branch first before this gets merged?

javerbukh:
I opened https://github.com/spacetelescope/jdaviz/pull/1398 as a companion to this work in jdaviz. It works well except for converting from wavelength to frequency - which works in the glue-jupyter POC notebook. Trying to use the data objects from the glue-jupyter notebook and then running `specviz.app.get_viewer("spectrum-viewer").state.x_display_unit = "Hz"` gives the following traceback:

<details>

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [8], in <cell line: 3>()
      1 # Y conversion works but X does not allow conversion from wavelength to frequency
----> 3 specviz.app.get_viewer("spectrum-viewer").state.x_display_unit = "Hz"
      4 specviz.app.get_viewer("spectrum-viewer").state.y_display_unit = "K"

File ~/opt/anaconda3/envs/glue_units/lib/python3.8/site-packages/echo/core.py:261, in HasCallbackProperties.__setattr__(self, attribute, value)
    260 def __setattr__(self, attribute, value):
--> 261     super(HasCallbackProperties, self).__setattr__(attribute, value)
    262     if self.is_callback_property(attribute):
    263         self._notify_global(**{attribute: value})

File ~/Documents/jdaviz_dev/jdaviz/jdaviz/core/freezable_state.py:13, in FreezableState.__setattr__(self, k, v)
     11 def __setattr__(self, k, v):
     12     if k[0] == '_' or k not in self._frozen_state:
---> 13         super().__setattr__(k, v)
     14     elif getattr(self, k) is None:
     15         # still allow Nones to be updated to initial values
     16         super().__setattr__(k, v)

File ~/opt/anaconda3/envs/glue_units/lib/python3.8/site-packages/echo/selection.py:37, in SelectionCallbackProperty.__set__(self, instance, value)
     35 if self.comparison_type == 'equality' or (self.comparison_type is None and np.isscalar(value)):
     36     if not any(value == x for x in choices):
---> 37         raise ValueError('value {0} is not in valid choices: {1}'.format(value, choices))
     38 if self.comparison_type == 'identity' or (self.comparison_type is None and not np.isscalar(value)):
     39     if not any(value is x for x in choices):

ValueError: value Hz is not in valid choices: ['m', 'usolRad', 'Ypc', 'AU', 'ypc', 'ZsolRad', 'klyr', 'lsec', 'Ylyr', 'YAU', 'earthRad', 'Tlyr', 'EAU', 'yAU', 'npc', 'Zm', 'Ym', 'dasolRad', 'Epc', 'dAU', 'Elyr', 'plyr', 'clyr', 'Mm', 'MsolRad', 'ksolRad', 'nm', 'MAU', 'nsolRad', 'dam', 'alyr', 'dsolRad', 'mm', 'YsolRad', 'am', 'asolRad', 'Gpc', 'ym', 'nlyr', 'aAU', 'Tpc', 'TsolRad', 'upc', 'ZAU', 'lyr', 'Zpc', 'pAU', 'PsolRad', 'cm', 'km', 'hpc', 'dapc', 'dlyr', 'msolRad', 'fAU', 'fm', 'zpc', 'GAU', 'um', 'hm', 'kAU', 'TAU', 'hlyr', 'cAU', 'EsolRad', 'hsolRad', 'jupiterRad', 'Gm', 'csolRad', 'fsolRad', 'mAU', 'Angstrom', 'zm', 'Em', 'dpc', 'mAngstrom', 'Ppc', 'psolRad', 'nAU', 'mpc', 'hAU', 'Pm', 'solRad', 'dalyr', 'zAU', 'Plyr', 'cpc', 'flyr', 'Zlyr', 'Glyr', 'daAU', 'dm', 'ysolRad', 'ppc', 'pc', 'Mpc', 'ulyr', 'pm', 'zlyr', 'uAU', 'fpc', 'zsolRad', 'PAU', 'GsolRad', 'ylyr', 'Tm', 'apc', 'Mlyr', 'mlyr', 'kpc']
```
</details>

Issue #312, opened 2022-05-18
Color widget misbehaves


jfoster17:
**Describe the bug**
The color widget (i.e. the box next to the eye icon that controls the color of the layer) in the layer options widget has two, potentially related, state bugs. 

1. The color widget initially shows up as white, rather than the color of the data layer. This does NOT happen if the initial color for a dataset is specified at viewer creation (e.g. `app.scatter2d(data='dataset1',x='x',y='y',color='red')` works just fine). 
2. If you drag in the color widget, instead of clicking distinctly, the color of the layer flickers between the old and the new color in an infinite loop.

**To Reproduce**
1. Create a scatter viewer
```python
import glue_jupyter as gj
from glue.core import Data
import numpy as np
app = gj.jglue()
dataset1 = Data(x=np.random.rand(100),y=np.random.rand(100),label='test')
app.add_data(dataset1 = dataset1)
app.scatter2d(data='dataset1',x='x',y='y')
```
2. Open up the layer options. The box showing the color of the layer should be gray, but it is white.3. 
3. Set the color by dragging in the color-picker widget and the color will flicker.4. 

**Expected behavior**
The color widget should start in sync with the viewer.layers[0].state.color
Dragging in the color picker should not break things.

**Video**

https://user-images.githubusercontent.com/3639698/169074229-d7e86477-f43a-4713-8dec-0f4a3513c759.mov

**Details:**

- Operating System: MacOS 12.3.1
- Python version Python 3.8.13
- Glue version 1.2.4
- glupyter-version 0.12.1.dev4+g254d624.d20220412 (latest main branch at the time)
- How you installed glue: pip

❯ jupyter --version
Selected Jupyter core packages...
IPython          : 8.2.0
ipykernel        : 6.12.1
ipywidgets       : 7.7.0
jupyter_client   : 7.2.2
jupyter_core     : 4.9.2
jupyter_server   : 1.16.0
jupyterlab       : 3.3.3
nbclient         : 0.5.13
nbconvert        : 6.4.5
nbformat         : 5.3.0
notebook         : 6.4.10
qtconsole        : 5.3.0
traitlets        : 5.1.1

**Additional context**
I have shown this for the bqplot scatter viewer, but it seems to impact the matplotlib viewers and the other bqplot viewers as well. I haven't been able to figure out how to get the state of this color-picker directly to write a test for this.

maartenbreddels:
@mariobuikhuizen maybe you want to take a look at this next week?

[...]

PR #313, opened 2022-05-24
fix: prevent color picker from entering an update loop

mariobuikhuizen:
## Description
This PR prevents the color picker from entering an update loop as described in #312 (2)

Fixes part 2 of #312

### DIFF
--- a/glue_jupyter/widgets/layer_options.py
+++ b/glue_jupyter/widgets/layer_options.py
@@ -8,6 +8,7 @@ import traitlets
 import ipywidgets as widgets
 from glue.core import message as msg
 from glue.core.hub import HubListener
+from ..utils import debounced
 from ..vuetify_helpers import WidgetCache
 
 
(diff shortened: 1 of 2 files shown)


maartenbreddels:
I wonder if we should use a trait instead of an event, to get less events. Also, does this solve the fundamental issue, or does it make it less likely to happen?

dhomeier:
@mariobuikhuizen does this still need further discussion (in particular, do you think there is still a way to implement @maartenbreddels's suggestion https://github.com/glue-viz/glue-jupyter/pull/284#issuecomment-1136046188)?

mariobuikhuizen:
We discovered this issue then: https://github.com/jupyter-widgets/ipywidgets/issues/3469, which is fixed in ipywidgets 7.7.1. @maartenbreddels suggestion in #284 should now work. I will finish this or that PR.

PR #315, opened 2022-07-12
Experimental image rotation

astrofrog:
Companion to https://github.com/glue-viz/glue/pull/2310 - need to actually avoid code duplication and so on, but this is just an experiment for now.

* [ ] Fix rotation of markers when overplotting on image


astrofrog:
I'm going to close this as it seems we are going to go with the front-end canvas rotation

PR #316, opened 2022-07-21
Fixed a bug that caused the image viewer zoom to be reset when adding additional datasets

astrofrog:
Ideally it would be good to understand why echo is calling the callbacks when the value doesn't change, but this fix will do for now.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+0.12.1 (unreleased)
+===================
+
+* Fixed a bug that caused the image viewer zoom to be reset when adding
+  additional datasets. [#316]
+
 0.12.0 (2022-04-07)
 ===================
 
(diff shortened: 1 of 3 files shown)


astrofrog:
If this works for jdaviz I should add a test before merging

svolfman:
Hello! I am a current SASP (Space Astronomy Summer Program) intern at STScI working on a Jdaviz software development project (https://github.com/spacetelescope/jdaviz/pull/1455) where I have witnessed the frustration associated with the viewer zoom resetting when adding and removing markers. 

I just wanted to let you know that I tested this bug fix in my local environment where I'm working on my project and the bug fix worked! Thank you!

dhomeier:
> Ideally it would be good to understand why echo is calling the callbacks when the value doesn't change, but this fix will do for now.

@astrofrog all I could track down was that it is the `reference_data` state that is triggering the callback, most likely from adding the links in
https://github.com/spacetelescope/jdaviz/blob/9e20b506201bfdbfb2820077dd45f82d9ea4faac/jdaviz/core/astrowidgets_api.py#L436-L439

but was not able to further isolate it to a specific `add_link` call. But see the comment on checking for already existing links; might perhaps be done in `glue.core.link_helpers.LinkSame`.

But I've also confirmed the that fix here works with the example notebook, so unless you have a new idea, would merge.

dhomeier:
Now released with 0.12.1 on PyPI.

pllim:
Thanks!

PR #317, opened 2022-09-13
fix: prevent color picker from entering an update loop

### LINKED ISSUE #312: Color widget misbehaves

maartenbreddels:
Supersedes #313 and #284
 * Closes #313
 * Closes #312 

We now make use of a trait, which automatically gives us a debounce, after the fix in [ipywidgets](https://github.com/jupyter-widgets/ipywidgets/issues/3469).

Also, we use the same trait for input and output (using v-model) instead of using a different trait (as in #313) or custom methods. Using a different trait for setting can cause loops or setting to old values when there is high latency.

### DIFF
--- a/glue_jupyter/widgets/layer_options.py
+++ b/glue_jupyter/widgets/layer_options.py
@@ -27,6 +27,7 @@ class LayerOptionsWidget(v.VuetifyTemplate, HubListener):
         self.viewer = viewer
 
         widgetCache = WidgetCache()
+        self.observe(self._update_glue_state_from_layers, 'layers')
 
         self.current_layers_data = None
 
@@ -90,11 +91,7 @@ class LayerOptionsWidget(v.VuetifyTemplate, HubListener):
         self.viewer.state.add_callback('layers', _update_layers_from_glue_state)
         _update_layers_from_glue_state()
 
-    def vue_toggle_visible(self, index):
-        state = self.viewer.layers[index].state
-        state.visible = not state.visible
-
-    def vue_set_color(self, data):
-        index = data['index']
-        color = data['color']
-        self.viewer.layers[index].state.color = color
+    def _update_glue_state_from_layers(self, *args):
+        for layer_data, layer in zip(self.layers, self.viewer.layers):
+            layer.state.visible = layer_data['visible']
+            layer.state.color = layer_data['color']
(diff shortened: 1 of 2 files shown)


astrofrog (approved):
This looks good to me!

PR #318, opened 2022-09-20
fix: do not upload float64 data to the widget, cast to float32

maartenbreddels:
WebGL cannot handle float64 data, so we need to cast it to float32. If not, the frontend will do it for us, but that will be a waste of bandwidth.

Addressses: https://github.com/spacetelescope/jdaviz/issues/1367


pllim:
Awesome! Is a release planned soon?

dhomeier:
Will do by the end of the week the latest (same for glue-astronomy)!

PR #319, opened 2022-09-20
BUG: Only recast float if image is not None

pllim:
This is a direct follow-up of #318 . Without this patch dev job is failing at Jdaviz.

### DIFF
--- a/glue_jupyter/bqplot/image/frb_mark.py
+++ b/glue_jupyter/bqplot/image/frb_mark.py
@@ -93,9 +93,9 @@ class FRBImage(ImageGL):
 
         # Get the array and assign it to the artist
         image = self.array_maker(bounds=bounds)
-        if image.dtype == np.dtype("float64"):
-            image = image.astype(np.float32)
         if image is not None:
+            if image.dtype == np.dtype("float64"):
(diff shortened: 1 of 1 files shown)


maartenbreddels:
Thanks. I wonder if it makes sense to run the jdaviz test suite here.

pllim:
> run the jdaviz test suite here

Hmm. Maybe @astrofrog has ideas for proper integration testing.

In the meantime, that is what the dev deps job in Jdaviz is for. 😸

dhomeier:
Thanks! I assume that is the only case where `image` might not have a `dtype`?

dhomeier:
> In the meantime, that is what the dev deps job in Jdaviz is for. 😸

Would still be helpful to have tests that can be run in an open PR here.

pllim:
I looked at the line changed here and it is not obvious to me what kind of test would actually trigger that line. There is no documentation. So I'll leave it to the experts to add a test. Thanks!

https://github.com/glue-viz/glue-jupyter/blob/ba222b6a9f6848f0ef4b6e1ff122825d95561044/glue_jupyter/bqplot/image/frb_mark.py#L18-L20

PR #320, opened 2022-10-07
Add a setting to control disabling of tools in other viewers

astrofrog:
By default, when selecting a tool, tools in other viewers are disabled. This can now be controlled with a setting:

```
app = gj.jglue(data=data, settings={'single_global_active_tool': False}) 
```


kecnry (approved):
works great for the use-case in jdaviz, thanks!

PR #324, opened 2022-10-11
fix: if image is not used in scatter layer, use an empty image.

maartenbreddels:
Fixes https://github.com/spacetelescope/jdaviz/issues/1664


astrofrog (approved):
Thanks!

pllim:
I think this got rid of the error. Thanks!

PR #325, opened 2022-10-11
Avoid sending any unecessary updates to the front-end

astrofrog:
This does two things:

* Fixes an infinite loop of callbacks in the layer options
* Avoids sending state updates to the front-end if the serialized JSON hasn't changed

@maartenbreddels - can you think of a better way to do the second of these? (the second commit in this PR)


pllim:
Seems to not have broken Imviz but I also didn't try very hard.

Also seems like things are "snappier" now (e.g., the contrast/bias sliders) but I could be imagining it. I do not have numbers to back this up.

PR #326, opened 2022-10-13
Fix IndexError that occurred in some cases during JSON to State translation

astrofrog:
The jdaviz developers have reported an issue with the JSON to Glue State translation which causes exceptions such as here:

https://github.com/spacetelescope/jdaviz/issues/438

The cause of this is that if the glue state has a ``CallbackList`` on a State with e.g. n elements and the front-end sends JSON which has only m elements (where m<n), this triggers an index error. The current fix in this PR is that if the JSON list has fewer items than the ``CallbackList``, we only update the first m elements in the CallbackList.

It's easier to explore this in a notebook so I have made one here:

https://gist.github.com/astrofrog/5fcc633bafb7fda7eee1d1eda192fd33

The bottom line is that there are two (I think mutually exclusive) options when the JSON updates a ``CallbackDict`` or ``CallbackList``:

* Either we require the length of lists in the JSON to match that of the ``CallbackList``, and for the keys in a dictionary to match or be a subset of those in a ``CallbackDict`` (which means that the length of a ``CallbackList`` and the keys of a ``CallbackDict`` cannot be changed by the front-end).
* Or we allow lists to have different lengths, which means that we shorten/lengthen ``CallbackList`` as needed, and we replace all keys in the ``CallbackDict`` with those in the JSON (meaning that ``{}`` in JSON would clear the ``CallbackDict``).

In other words, do we allow the front-end to modify the structure of the ``CallbackDict`` and ``CallbackList``, or only the items inside them?

This PR currently follows the first option - we were basically doing this modulo a bug. With the current implementation:

* If the JSON list is shorter than the CallbackList only the items in the JSON list are changed in the CallbackList
* If the JSON list is longer than the CallbackList the extra elements are ignored
* The JSON list can actually be specified also as a dictionary of index: value pairs in cases where the front-end wants to only update some items in the list
* Keys in the JSON dict that don't exist in ``CallbackDict`` are ignored

Are there cases where we think this would not be sufficient and where we would need the front-end to actually modify the number of items in a CallbackList or a CallbackDict?

### DIFF
--- a/glue_jupyter/state_traitlets_helpers.py
+++ b/glue_jupyter/state_traitlets_helpers.py
@@ -45,12 +45,14 @@ def update_state_from_dict(state, changes):
             if isinstance(getattr(state, name), CallbackList):
                 callback_list = getattr(state, name)
                 for i in range(len(callback_list)):
-                    if isinstance(callback_list[i], State):
-                        update_state_from_dict(callback_list[i], changes[name][i])
-                    else:
-                        if (changes[name][i] != MAGIC_IGNORE and
-                                callback_list[i] != changes[name][i]):
-                            callback_list[i] = changes[name][i]
+                    if (isinstance(changes[name], dict) and i in changes[name]
+                            or isinstance(changes[name], list) and i < len(changes[name])):
+                        if isinstance(callback_list[i], State):
+                            update_state_from_dict(callback_list[i], changes[name][i])
+                        else:
+                            if (changes[name][i] != MAGIC_IGNORE and
+                                    callback_list[i] != changes[name][i]):
+                                callback_list[i] = changes[name][i]
             elif isinstance(getattr(state, name), CallbackDict):
                 callback_dict = getattr(state, name)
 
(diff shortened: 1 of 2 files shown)


rosteen:
I confirmed that this fixes the error in the case I've been using to reproduce it, thanks!

glue_jupyter/state_traitlets_helpers.py
```diff
@@ -45,12 +45,14 @@ def update_state_from_dict(state, changes):
             if isinstance(getattr(state, name), CallbackList):
                 callback_list = getattr(state, name)
                 for i in range(len(callback_list)):
-                    if isinstance(callback_list[i], State):
-                        update_state_from_dict(callback_list[i], changes[name][i])
-                    else:
-                        if (changes[name][i] != MAGIC_IGNORE and
-                                callback_list[i] != changes[name][i]):
-                            callback_list[i] = changes[name][i]
+                    if (isinstance(changes[name], dict) and i in changes[name]
```
dhomeier, line 48: `changes[name]` can only be an index-dict here (as in the test), right? Should perhaps be noted somewhere for clarity.

astrofrog:
I'm going to go ahead and merge this since it fixes the existing error in jdaviz - we can always revisit in future if we want to change the behavior and allow the front-end to be able to modify list/dictionary lengths.

glue_jupyter/state_traitlets_helpers.py
astrofrog: Done

PR #332, opened 2022-10-31
MNT: Pin maxversion of glue-core until package catches up

pllim:
## Description

@kecnry said glue-jupyter will be incompatible with glue-core when glue-core v1.6 is released.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -14,7 +14,7 @@ python_requires = >=3.7
 setup_requires =
   setuptools_scm
 install_requires =
-    glue-core>=1.3.0
+    glue-core>=1.3.0,<1.6
     glue-vispy-viewers>=1.0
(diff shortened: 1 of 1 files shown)


astrofrog:
This should still work with older versions of glue-core, so I will close this

pllim:
I thought @kecnry said otherwise...?

kecnry:
I said the opposite, previous version of glue-jupyter (0.14.1) would show deprecation warnings for the new version (1.6) of glue.  With 0.14.2 released now, I don't think there's anything we can/should do here.  If anyone updates glue but not glue-jupyter, they might see deprecation warnings and some odd behavior, which can be resolved by also updating glue-jupyter to latest.

pllim:
OK, I must have gotten confused. Sorry for the noise!

PR #334, opened 2022-11-09
Allow adjusting z-order for scatter and histogram layers

Carifio24:
This PR adds support for modifying the z-orders of layers in the scatter and histogram viewers. In the layer artist, this is achieved by re-ordering all of the marks of viewer's figure, sorted by layer z-order. In the UI, since we don't (yet) have anywhere to do the drag-and-drop reordering that one would do in desktop glue, this PR also adds `IntText` widgets to the layer option widgets for the histogram and scatter layers for setting the z-order. Note that in principle there's nothing that prevents a similar approach for the profile and image viewers, aside from the fact that the `IntText` widget doesn't match the Vuetify styling of their layer option widgets.

The last piece of this makes some changes to the layer options widget. Currently the layer options widget stores the index of the selected layer in the layers list, and displays the panel that corresponds to that index in the list of layers. Without any adjustment, this leads to an unpleasant effect where the selected layer will change if a user changes the z-order such that layers get re-ordered. To work around this, I've addeds member value that stores the information for the current layer and panel. When the layers get updated from glue, there's a check of whether the current layer info exists in the layer list (which it should if the change is a z-order change). If so, selected is updated to match the new index, and so the user's current layer is unchanged. Otherwise there is no change from the current behavior.

### DIFF
--- a/glue_jupyter/bqplot/histogram/layer_artist.py
+++ b/glue_jupyter/bqplot/histogram/layer_artist.py
@@ -112,7 +112,9 @@ class BqplotHistogramLayerArtist(LayerArtist):
 
         if not self.enabled:
             return
-        # TODO: set visual attrs
+        # TODO: set other visual attrs
+        sorted_layers = sorted(self.view.layers, key=lambda layer: layer.state.zorder)
+        self.view.figure.marks = [layer.bars for layer in sorted_layers]
         self.redraw()
 
     def _update_histogram(self, force=False, **kwargs):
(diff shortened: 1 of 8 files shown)


dhomeier:
Seems tests here have failed with an even older conflict of jupyter-core and jupyter-client; hopefully rebasing should fix this as well.

dhomeier:
codecov is complaining about `_update_zorder`, any ideas how to test that?

Carifio24:
@dhomeier I've added a test for the histogram and scatter z-ordering, which checks that the relevant layer marks are in the correct (relative) order in the figure.

dhomeier (approved):
Nicely done, thanks!

PR #337, opened 2022-11-16
Set row number column width in Table widget dynamically from no. of digits

kecnry:
## Description

This pull request modifies the width of the first column (row number) in the table widget to have its width calculated based on the space needed for the largest index, by assuming each digit requires 20px.

Before:

Screen Shot 2022-11-16 at 12 30 54 PM

After:

Screen Shot 2022-11-16 at 12 29 10 PM

### DIFF
--- a/glue_jupyter/table/table.vue
+++ b/glue_jupyter/table/table.vue
@@ -15,7 +15,7 @@
       <template v-slot:header="props">
         <thead>
           <tr>
-            <th style="padding: 0 10px; width: 40px">#</th>
+            <th :style="'padding: 0 10px; width: '+Math.ceil(Math.log10(total_length))*20+'px'">#</th>
             <th style="padding: 0 1px; width: 30px" v-if="selection_enabled">
               <v-btn icon color="primary" text small @click="apply_filter">
                 <v-icon>filter_list</v-icon>
--- a/glue_jupyter/table/table.vue
+++ b/glue_jupyter/table/table.vue
@@ -15,7 +15,7 @@
       <template v-slot:header="props">
         <thead>
           <tr>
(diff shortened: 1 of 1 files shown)


glue_jupyter/table/table.vue
```diff
@@ -15,7 +15,7 @@
       <template v-slot:header="props">
         <thead>
           <tr>
-            <th style="padding: 0 10px; width: 40px">#</th>
+            <th :style="'padding: 0 10px; width: '+Math.ceil(Math.log10(total_length))*20+'px'">#</th>
```
dhomeier, line 18: Quite a bit of Math here! I am well out of my depth with the syntax, but I assume there is no equivalent of
`len(f"{total_length}")`? 😆

kecnry: `total_length` is an integer of how many entries are in the table.  So `ceil(log10(total_length))` is intended to find the maximum number of digits needed, by mapping 1 to 0 digits (perhaps we should have an exception to force this to 1?), 2-10 to 1 digit, 11-100 to 2 digits, etc, which with the exception of `total_length==1` is what would be expected since the table is zero-indexed.

dhomeier: Yes, it's clear that the method works, I was just a little lost in the remaining vuetify template code around it.
It might certainly be done the Python way as well, by adding a `max_width` to `TableBase`, but maybe not worth it as it would need to be updated as well.

dhomeier, line 18:
```suggestion
            <th :style="'padding: 0 10px; width: '+Math.max(1, Math.ceil(Math.log10(total_length)))*20+'px'">#</th>
```
Catching case `1` is a good idea indeed.

dhomeier (approved):
Thanks, that's a very helpful improvement!

PR #338, opened 2022-12-07
Fix bug where empty histogram layer does not redraw

Carifio24:
This PR aims to address the same issue that https://github.com/glue-viz/glue/pull/2300 dealt with in Qt glue, which is that if a subset group is modified to make it empty, the corresponding histogram layer artist is not redrawn. The changes here follow essentially the same logic as in the Qt update. Note that this PR only needs to update the layer artist, since the bqplot and Qt histogram viewers use the same viewer state class.

### DIFF
--- a/glue_jupyter/bqplot/histogram/layer_artist.py
+++ b/glue_jupyter/bqplot/histogram/layer_artist.py
@@ -58,7 +58,7 @@ class BqplotHistogramLayerArtist(LayerArtist):
         if self.bins is None:
             return  # can happen when the subset is empty
 
-        if self.bins.size == 0 or self.hist_unscaled.sum() == 0:
+        if self.bins.size == 0:
             return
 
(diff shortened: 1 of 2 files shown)


dhomeier:
Thanks, at a first glance this looks sensible.
Could you rebase please, that should hopefully get the tests to run?

Carifio24:
@dhomeier Just rebased and CI is looking good now, with the exception of codecov. There is a new test in this PR that checks the specific redrawing issue, but it doesn't touch all of the changes.

dhomeier:
It's missing a case where all histogram bins are empty?

Carifio24:
Some of the changes were to account for redrawing an empty layer when in cumulative/normalized/log modes. I've updated the test to make sure that the layer is still empty in a variety of viewer configurations.

dhomeier (approved):
Perfect on coverage, thanks!

PR #346, opened 2023-02-20
Implement support for showing world labels in profile viewer when in pixel space

astrofrog:
This is a feature that has been requested for jdaviz but is a generically useful feature - similar to what we do for images where we show images in pixel space and show the tick values/labels in world space, this PR implements something similar for the profile viewer in glue-jupyter (and we could in principle also extend this to the image viewer in glue-jupyter too).

Screenshot from 2023-02-20 11-53-51

Currently this needs https://github.com/bqplot/bqplot/pull/1541 rebased on the bqplot v0.12.x branch.

This is very incomplete but already mostly works with a simple example (notebook included). The main things to still do are:

* [ ] Investigate and fix issue with labels sometimes updating sometimes not (race condition in bqplot?). More details in a follow-up comment below
* [ ] Support cases where the data has more than one dimension
* [ ] Properly deal with display units (for now just works with original WCS units)
* [ ] Determine how to do 'pretty' exponential notation
* [ ] Add a callback property on the state to determine whether or not to show world coordinates when in pixel space
* [ ] Add tests

### DIFF
--- a/glue_jupyter/bqplot/profile/viewer.py
+++ b/glue_jupyter/bqplot/profile/viewer.py
@@ -9,6 +9,10 @@ from ..common.viewer import BqplotBaseView
 
 from .layer_artist import BqplotProfileLayerArtist
 
+from astropy import units as u
+from astropy.visualization.wcsaxes.formatter_locator import ScalarFormatterLocator
+
+from glue.core.component_id import PixelComponentID
 from glue_jupyter.common.state_widgets.layer_profile import ProfileLayerStateWidget
 from glue_jupyter.common.state_widgets.viewer_profile import ProfileViewerStateWidget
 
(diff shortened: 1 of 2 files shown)


astrofrog:
@maartenbreddels - a couple of questions for you:

1. The labels sometimes update and sometimes don't - could you take a look at what might be happening? (just run the notebook here and pan/zoom x-axis to reproduce). Note that I am already using https://github.com/bqplot/bqplot/pull/1541 so there must be some further issues? See following animation:

[Screencast from 20-02-23 11:56:54.webm](https://user-images.githubusercontent.com/314716/220099993-bea7552a-34f4-40dd-bb36-8fbf167d54e8.webm)

See the final state of the video where some labels seem to be updated and some not.

2. Is there a way to have 'pretty' exponential notation for labels? (kind of like 2 x 10^3 but with the superscript actually being a superscript. I noticed that I can set tick_labels to be values instead of strings which then lets the front-end format it, but I end up with values like 10000000000. Is there a way to tell bqplot to be smarter about labelling this?

PR #347, opened 2023-04-03
Allow cmap and size to take cateogricals

jfoster17:
# Allow Linear scaling on cmap and size from categorical component

## Description

This just calls ensure_numerical on components that are passed to cmap and size. Required if we allow categoricals to set these scalings (as in https://github.com/glue-viz/glue/pull/2317), but a fine idea anyway.


jfoster17:
This is good-to-go (it passes tests when rebased on #348). Sorry for the visual spam in re-ordering the imports, I was configured to run isort automatically and... it is a little helpful to have the imports in a rational order :)

PR #348, opened 2023-04-03
Remove pkg resources call

jfoster17:
# Remove call to pkg_resources to work with latest setuptools

## Description

Much like in https://github.com/glue-viz/glue/pull/2365 and https://github.com/glue-viz/echo/pull/36, this removes a deprecated call to pkg_resources. Note since glue-jupyter requires python>3.8 and we're just getting the version info (a simple call), we don't have to rely on the importlib backport packages at all.

### DIFF
--- a/glue_jupyter/__init__.py
+++ b/glue_jupyter/__init__.py
@@ -1,6 +1,6 @@
 from __future__ import absolute_import
 
-from pkg_resources import get_distribution, DistributionNotFound
+import importlib.metadata
 
 from IPython.display import display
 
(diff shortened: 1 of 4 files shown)


astrofrog:
@jfoster17 - it looks like the first commit is unrelated to this PR?

jfoster17:
I think this coverage failure is just because my commit here reduced the number of lines in a file with less than 100% coverage.

I'm ignoring some DepreciationWarnings here that were coming from other glue plug-ins (specifically `glue-vispy-viewers`) and `mpl_toolkits`. Hopefully we can remove these once everything is updated. 

@astrofrog -- I think this is ready to go now.

PR #356, opened 2023-06-06
FEAT: Annulus draw tool

pllim:
## Description

We need this for Imviz annulus support. This blocks:

* https://github.com/spacetelescope/jdaviz/pull/2240

cc @dhomeier and @astrofrog

### DIFF
--- a/glue_jupyter/bqplot/common/tools.py
+++ b/glue_jupyter/bqplot/common/tools.py
@@ -1,14 +1,23 @@
+import os
 from contextlib import nullcontext
 
-import os
+import numpy as np
 from bqplot import PanZoom
 from bqplot.interacts import BrushSelector, BrushIntervalSelector
 from bqplot_image_gl.interacts import BrushEllipseSelector
+from glue import __version__ as glue_version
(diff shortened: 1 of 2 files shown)


pllim:
p.s. I am not good with the icon stuff, so if you have someone over at Glueviz that makes the proper icons, please feel free to replace the file directly on my branch or open a PR to this PR. Thanks!

pllim:
> icon looks strange

I am not surprised if it looks strange. I don't have time or interest to make pretty icon especially since Jdaviz uses its own icons anyway and we have a professional designer doing that for us. So, if you have a prettier version, please replace the one in this PR.

> I wonder whether after the first click to select, the user should be able to then continue moving the cursor to edit the inner radius?

That could be a follow-up PR. Would be nice to be able to do that at draw time, yes, but I am just not sure what kind of click and when would be the most intuitive, so definitely need some iterations.

astrofrog:
I have fixed the icon:

Screenshot 2023-06-16 at 12 07 45

astrofrog (approved):
Thanks!

PR #357, opened 2023-06-08
Fix duplicate xrange region bug when operators are used on existing regions

pllim:
@javerbukh had been battling duplicate sub-regions appearing in his work over at https://github.com/spacetelescope/jdaviz/pull/2237 .

Here is a workflow to reproduce the bug:

1. Draw a xrange region.
2. Select the Subset generated from Step 1.
3. Select XOR tool.
4. Select another region. This would turn Subset 1 into composite subset.

What I noticed with this bug is, the composition went a level deeper than it should have:

```
AND
|__ AND
|   |__ Range
|   |__ INV
|       |__ Range
|__ INV
    |__ Range
```

When it should be much simpler like this:

```
AND
|__ Range
|__ INV
    |__ Range
```

Since we never heard back from SMEs about this, I decided to have a look. I asked myself why are we not seeing this on spatial subsets but it appears always on spectral subset? Then I noticed that the tools logic for spatial region has more stuff to it than the xrange tool, so I decided to copy stuff over, and that appears to have fixed the bug, but I would need @javerbukh to confirm for sure.

Fix https://github.com/spacetelescope/jdaviz/issues/2061

[🐱](https://jira.stsci.edu/browse/JDAT-3138)

### DIFF
--- a/glue_jupyter/bqplot/profile/tests/test_viewer.py
+++ b/glue_jupyter/bqplot/profile/tests/test_viewer.py
@@ -1,12 +1,15 @@
 from astropy.wcs import WCS
 
 import numpy as np
+import pytest
 from numpy.testing import assert_allclose, assert_equal
 
 from astropy import units as u
 
 from glue.core import Data
+from glue.core.edit_subset_mode import AndNotMode
 from glue.core.roi import XRangeROI
+from glue.core.subset import AndState, InvertState, RangeSubsetState
 from glue.config import unit_converter, settings
 from glue.plugins.wcs_autolinking.wcs_autolinking import WCSLink
 
@@ -61,6 +64,7 @@ class SpectralUnitConverter:
                                                           equivalencies=u.spectral())
 
 
+@pytest.mark.filterwarnings("ignore:No observer defined on WCS.*")
 def test_unit_conversion(app):
 
     settings.UNIT_CONVERTER = 'test-spectral'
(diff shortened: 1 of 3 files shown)


javerbukh:
This fixes the duplicating range subregions issue and this issue https://github.com/spacetelescope/jdaviz/issues/2061 thank you!!

pllim:
Maybe I should add test...

pllim:
@javerbukh , does this break any workaround in `jdaviz`?

javerbukh:
@pllim It exposes a bug but does not create one. I will fix it on the jdaviz side. This PR looks good to me!

glue_jupyter/bqplot/profile/viewer.py
```diff
@@ -28,7 +28,7 @@ class BqplotProfileView(BqplotBaseView):
     _layer_style_widget_cls = ProfileLayerStateWidget
 
     tools = ['bqplot:home', 'bqplot:panzoom', 'bqplot:panzoom_x', 'bqplot:panzoom_y',
-             'bqplot:xrange']
+             'bqplot:xrange', 'bqplot:yrange']
```
pllim, line 31: I don't see why you don't want this? I mainly added this here so I can test it, but if there is a way for me to add this just for the test and not here and you really don't like here, then please let me know.

glue_jupyter/bqplot/profile/tests/test_viewer.py
```diff
@@ -157,3 +161,80 @@ def test_unit_conversion(app):
...
+    session.edit_subset_mode.mode = AndNotMode
+
+    tool.interact.brushing = True
+    tool.interact.selected = [25, 45]
+    tool.interact.brushing = False
+    tool.deactivate()
+
+    sbst = session.data_collection.subset_groups[0].subset_state
+    assert isinstance(sbst, AndState)
+    assert isinstance(sbst.state1, RangeSubsetState)
```
pllim, line 199: Without this patch, the two new tests would fail like this (I ran it on current `main`):

```
>       assert isinstance(sbst.state1, RangeSubsetState)
E       assert False
E        +  where False = isinstance(<glue.core.subset.AndState ...>, RangeSubsetState)
E        +    where <glue.core.subset.AndState ...> = <glue.core.subset.AndState ...>.state1
```

But the tests should pass here.

glue_jupyter/bqplot/profile/tests/test_viewer.py
```diff
@@ -157,3 +161,80 @@ def test_unit_conversion(app):
...
+
+    d2 = Data(f2=[2000, 1000, 3000])
+    d2.get_component('f2').units = 'mJy'
+    d2.coords = wcs2
+
+    data_collection = session.data_collection
+    data_collection.append(d2)
+
+    viewer = app.profile1d(data=d2)
+    tool = viewer.toolbar.tools['bqplot:xrange']
```
pllim, line 184: We need to test the tool with brushing because `viewer.apply_roi(...)` will not trigger the bug.

glue_jupyter/bqplot/profile/viewer.py
astrofrog: No this seems fine I think

astrofrog (approved):
This looks good to me, thanks!

PR #362, opened 2023-06-23
Modify visual test to check visual test infrastructure

astrofrog:
This is to check that the test fails and that giles works


astrofrog:
Seems to be working! Thanks for the help with giles @Cadair!

PR #365, opened 2023-06-27
Add giles config to pyproject.toml

astrofrog:
Hopefully giles will be able to start reporting


astrofrog:
Ah I forgot that the config is read from main so I need to merge this before trying it out

PR #366, opened 2023-06-28
Feat: Solara compatible

maartenbreddels:
We cannot re-use the same widget with multiple connections/users in Solara,

Should fix https://github.com/glue-viz/glue-jupyter/pull/363

### DIFF
--- a/glue_jupyter/vuetify_layout.py
+++ b/glue_jupyter/vuetify_layout.py
@@ -3,12 +3,11 @@
 # default layout.
 
 import ipyvuetify as v
-import traitlets
 import ipywidgets as widgets
+import traitlets
 from ipywidgets import widget_serialization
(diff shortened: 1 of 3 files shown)


maartenbreddels:
@astrofrog @mariobuikhuizen I wonder if there should be a way to create a widget based on a custom key (e.g. in this case per icon, and in the case of solara also per connection) as a replacement for globally shared widgets that do not work in solara.

A feature/function like this could be provided by solara, but that requires a dependency of glue-jupyter on solara.

Solara's [memoize](https://solara.dev/api/memoize) function is close to this but is not per connection by default.

astrofrog:
I'm going to push an attempted fix for the CI failures

astrofrog (approved):

PR #370, opened 2023-07-05
Circle draw tool must produce circle

pllim:
User was confused on why they get back elliptical region when they used the circular draw tool on their image. You can basically reproduce this behavior using https://github.com/spacetelescope/jdaviz/issues/697 even though I think @camipacifici reported a different workflow.

With this patch, even though the circle tool obviously produced something that does not look circular in Lab, it still gives a circle.

Screenshot 2023-07-05 173720

~I don't see the affected code tested anywhere in this repo, so I am unable to add new tests.~

[🐱](https://jira.stsci.edu/browse/JDAT-3545)

### DIFF
--- a/glue_jupyter/bqplot/common/tools.py
+++ b/glue_jupyter/bqplot/common/tools.py
@@ -218,34 +218,26 @@ class BqplotCircleMode(BqplotSelectionTool):
             if self.interact.selected_x is not None and self.interact.selected_y is not None:
                 x = self.interact.selected_x
                 y = self.interact.selected_y
-                # similar to https://github.com/glue-viz/glue/blob/b14ccffac6a5
-                # 271c2869ead9a562a2e66232e397/glue/core/roi.py#L1275-L1297
-                # We should now check if the radius in data coordinates is the same
-                # along x and y, as if so then we can return a circle, otherwise we
-                # should return an ellipse.
+                # We enforce strict circle.
                 xc = x.mean()
                 yc = y.mean()
                 rx = abs(x[1] - x[0])/2
                 ry = abs(y[1] - y[0])/2
-                # We use a tolerance of 1e-2 below to match the tolerance set in glue-core
-                # https://github.com/glue-viz/glue/blob/6b968b352bc5ad68b95ad5e3bb25550782a69ee8/glue/viewers/matplotlib/state.py#L198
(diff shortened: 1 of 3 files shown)


pllim:
I don't think the `pydantic.errors.PydanticImportError: `BaseSettings` has been moved to the `pydantic-settings` package.` error in CircleCI is related.

pllim:
p.s. I tried having both tools subclass from a common base class but that broke the click-and-drag feature for some reason.

astrofrog:
Thanks! This approach doesn't quite work because if I make a simple scatter plot in glue-jupyter where the two axes might be quantities with different scales (e.g. x is velocity between -1 and 1 and y is position between -100 and 100) the current 'circle' tool will produce weird results - e.g.

Screenshot 2023-07-10 at 13 45 48

I think a cleaner solution would be to keep the current circular tool as is but add a class attribute e.g. ``_strict_circle`` that defaults to ``False`` and check that attribute where we currently construct a ``CircularROI``. You can then make a 'strict circle' subclass that simply sets ``_strict_circle = True``, and use ``bqplot:strictcircle`` in Imviz.

glue_jupyter/bqplot/common/tools.py
```diff
@@ -218,34 +218,26 @@ def update_selection(self, *args):
...
                 yc = y.mean()
                 rx = abs(x[1] - x[0])/2
                 ry = abs(y[1] - y[0])/2
-                # We use a tolerance of 1e-2 below to match the tolerance set in glue-core
-                # https://github.com/glue-viz/glue/blob/6b968b352bc5ad68b95ad5e3bb25550782a69ee8/glue/viewers/matplotlib/state.py#L198
-                if np.allclose(rx, ry, rtol=1e-2):
-                    roi = CircularROI(xc=xc, yc=yc, radius=rx)
-                else:
-                    roi = EllipticalROI(xc=xc, yc=yc, radius_x=rx, radius_y=ry)
+                r = (rx + ry) * 0.5
```
dhomeier, line 226:
```suggestion
                r = np.sqrt((rx**2 + ry**2) * 0.5)
```
adds a little computational overhead, but I think using arithmetic mean here could create a circle passing somewhat besides the actual pointer position

pllim:
This has been superseded by #376

PR #375, opened 2023-07-18
Restore, but deprecate BqplotScatterLayerState

dhomeier:
## Description

Removing `BqplotScatterLayerState` from `glue_jupyter.bqplot.scatter.layer_artist` broke an application in Jdaviz:
https://github.com/glue-viz/glue-jupyter/pull/363#discussion_r1262684840

This is restoring the class; @astrofrog is it OK to just revert that change, or do you have an idea for a fix that does not need `echo`?

### DIFF
--- a/glue_jupyter/bqplot/scatter/layer_artist.py
+++ b/glue_jupyter/bqplot/scatter/layer_artist.py
@@ -1,5 +1,6 @@
 import bqplot
 import numpy as np
+import warnings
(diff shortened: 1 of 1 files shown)


dhomeier:
Don't know if my earlier comment disappeared into thin air now – should that also be marked deprecated for later removal then?

astrofrog:
Yes maybe that would be best - thanks!

astrofrog (approved):

PR #376, opened 2023-07-18
Allow enforcing exactly circular ROI in draw tool 

dhomeier:
## Description

Substitute for #370 that does not alter the default behaviour (of drawing an ellipse when x and y radius are > 2 % different), to avoid issues as https://github.com/glue-viz/glue-jupyter/pull/370#issuecomment-1628892239.

Instead returning a `CircularROI` is only enforced by setting a kwarg `strict_circle=True`, which is made available as default in `BqplotTrueCircleMode` and `'bqplot:truecircle'`.

@camipacifici, @pllim I checked with the tests adapted from #370, but could not test the Jdaviz examples, since I cannot reproduce the original issue with either workflow at the moment. But replacing `BqplotCircleMode` with `BqplotTrueCircleMode` in Jdaviz tools with this PR should in hopefully fix them.

@astrofrog yet to decide if this is to be exposed by a public kwarg; also not sure if we still need the additional interface (and test) through `'bqplot:ellipse'` with this approach.

### DIFF
--- a/glue_jupyter/bqplot/common/tools.py
+++ b/glue_jupyter/bqplot/common/tools.py
@@ -275,7 +275,7 @@ class BqplotCircleMode(BqplotSelectionTool):
     action_text = 'Circular ROI'
     tool_tip = 'Define a circular region of interest'
 
-    def __init__(self, viewer, roi=None, finalize_callback=None, **kwargs):
+    def __init__(self, viewer, roi=None, finalize_callback=None, strict_circle=False, **kwargs):
 
         super().__init__(viewer, **kwargs)
 
@@ -299,6 +299,7 @@ class BqplotCircleMode(BqplotSelectionTool):
         self.interact.observe(self.on_selection_change, "selected_x")
         self.interact.observe(self.on_selection_change, "selected_y")
         self.finalize_callback = finalize_callback
+        self._strict_circle = strict_circle
 
     def update_selection(self, *args):
         if self.interact.brushing:
@@ -309,17 +310,19 @@ class BqplotCircleMode(BqplotSelectionTool):
                 y = self.interact.selected_y
                 # similar to https://github.com/glue-viz/glue/blob/b14ccffac6a5
                 # 271c2869ead9a562a2e66232e397/glue/core/roi.py#L1275-L1297
-                # We should now check if the radius in data coordinates is the same
-                # along x and y, as if so then we can return a circle, otherwise we
-                # should return an ellipse.
+                # If _strict_circle set, enforce returning a circle; otherwise check
+                # if the radius in data coordinates is (nearly) the same along x and y,
+                # to return a circle as well, else we should return an ellipse.
                 xc = x.mean()
                 yc = y.mean()
                 rx = abs(x[1] - x[0])/2
                 ry = abs(y[1] - y[0])/2
                 # We use a tolerance of 1e-2 below to match the tolerance set in glue-core
                 # https://github.com/glue-viz/glue/blob/6b968b352bc5ad68b95ad5e3bb25550782a69ee8/glue/viewers/matplotlib/state.py#L198
-                if np.allclose(rx, ry, rtol=1e-2):
-                    roi = CircularROI(xc=xc, yc=yc, radius=rx)
+                if self._strict_circle:
+                    roi = CircularROI(xc=xc, yc=yc, radius=np.sqrt((rx**2 + ry**2) * 0.5))
+                elif np.allclose(rx, ry, rtol=1e-2):
+                    roi = CircularROI(xc=xc, yc=yc, radius=(rx + ry) * 0.5)
                 else:
                     roi = EllipticalROI(xc=xc, yc=yc, radius_x=rx, radius_y=ry)
                 self.viewer.apply_roi(roi)
@@ -330,7 +333,10 @@ class BqplotCircleMode(BqplotSelectionTool):
         if isinstance(roi, CircularROI):
             rx = ry = roi.radius
         elif isinstance(roi, EllipticalROI):
-            rx, ry = roi.radius_x, roi.radius_y
+            if self._strict_circle:
+                rx, ry = np.sqrt((roi.radius_x ** 2 + roi.radius_y ** 2) * 0.5)
+            else:
+                rx, ry = roi.radius_x, roi.radius_y
         else:
             raise TypeError(f'Cannot initialize a BqplotCircleMode from a {type(roi)}')
         self.interact.selected_x = [roi.xc - rx, roi.xc + rx]
@@ -346,6 +352,21 @@ class BqplotCircleMode(BqplotSelectionTool):
             self.interact.selected_x = None
             self.interact.selected_y = None
         super().activate()
+
+
+@viewer_tool
+class BqplotTrueCircleMode(BqplotCircleMode):
+
+    tool_id = 'bqplot:truecircle'
+    tool_tip = 'Define a strictly circular region of interest'
+
+    def __init__(self, viewer, roi=None, finalize_callback=None, **kwargs):
+
+        super().__init__(viewer, **kwargs)
+
+        if not kwargs.pop('strict_circle', True):
+            raise ValueError('BqplotTrueCircleMode cannot have strict_circle=False')
+        self._strict_circle = True
 
 
 @viewer_tool
@@ -582,7 +603,10 @@ class ROIClickAndDrag(InteractCheckableTool):
             if layer.visible and isinstance(subset_state, RoiSubsetState):
                 roi = subset_state.roi
                 if roi.defined() and roi.contains(x, y):
-                    if isinstance(roi, (EllipticalROI, CircularROI)):
+                    if isinstance(roi, EllipticalROI):
+                        self._active_tool = BqplotEllipseMode(
+                            self.viewer, roi=roi, finalize_callback=self.release)
+                    elif isinstance(roi, CircularROI):
                         self._active_tool = BqplotCircleMode(
                             self.viewer, roi=roi, finalize_callback=self.release)
                     elif isinstance(roi, (PolygonalROI, RectangularROI)):
--- a/glue_jupyter/bqplot/common/tools.py
+++ b/glue_jupyter/bqplot/common/tools.py
@@ -279,7 +279,7 @@ class BqplotCircleMode(BqplotSelectionTool):
     action_text = 'Circular ROI'
     tool_tip = 'Define a circular region of interest'
 
-    def __init__(self, viewer, roi=None, finalize_callback=None, strict_circle=None, **kwargs):
+    def __init__(self, viewer, roi=None, finalize_callback=None, **kwargs):
 
         super().__init__(viewer, **kwargs)
 
@@ -295,7 +295,7 @@ class BqplotCircleMode(BqplotSelectionTool):
         border_style['stroke'] = INTERACT_COLOR
         self.interact.style = style
         self.interact.border_style = border_style
-        self._strict_circle = strict_circle
+        self._strict_circle = False
 
         if roi is not None:
             self.update_from_roi(roi)
@@ -336,7 +336,7 @@ class BqplotCircleMode(BqplotSelectionTool):
     def update_from_roi(self, roi):
         if isinstance(roi, CircularROI):
             rx = ry = roi.radius
-            if isinstance(roi, TrueCircularROI) and self._strict_circle is not False:
+            if isinstance(roi, TrueCircularROI):
                 self._strict_circle = True
         elif isinstance(roi, EllipticalROI):
             if self._strict_circle:
@@ -369,9 +369,6 @@ class BqplotTrueCircleMode(BqplotCircleMode):
     def __init__(self, viewer, roi=None, finalize_callback=None, **kwargs):
 
         super().__init__(viewer, **kwargs)
-
-        if kwargs.pop('strict_circle', True) is False:
-            raise ValueError('BqplotTrueCircleMode cannot have strict_circle=False')
         self._strict_circle = True
 
 
--- a/glue_jupyter/bqplot/common/tools.py
+++ b/glue_jupyter/bqplot/common/tools.py
@@ -27,6 +27,10 @@ INTERACT_COLOR = '#cbcbcb'
 ICONS_DIR = os.path.join(os.path.dirname(__file__), '..', '..', 'icons')
 
 
+class TrueCircularROI(CircularROI):
+    pass
+
+
 class InteractCheckableTool(CheckableTool):
 
     def __init__(self, viewer):
@@ -275,7 +279,7 @@ class BqplotCircleMode(BqplotSelectionTool):
     action_text = 'Circular ROI'
     tool_tip = 'Define a circular region of interest'
 
-    def __init__(self, viewer, roi=None, finalize_callback=None, strict_circle=False, **kwargs):
+    def __init__(self, viewer, roi=None, finalize_callback=None, strict_circle=None, **kwargs):
 
         super().__init__(viewer, **kwargs)
 
@@ -291,6 +295,7 @@ class BqplotCircleMode(BqplotSelectionTool):
         border_style['stroke'] = INTERACT_COLOR
         self.interact.style = style
         self.interact.border_style = border_style
+        self._strict_circle = strict_circle
 
         if roi is not None:
             self.update_from_roi(roi)
@@ -299,7 +304,6 @@ class BqplotCircleMode(BqplotSelectionTool):
         self.interact.observe(self.on_selection_change, "selected_x")
         self.interact.observe(self.on_selection_change, "selected_y")
         self.finalize_callback = finalize_callback
-        self._strict_circle = strict_circle
 
     def update_selection(self, *args):
         if self.interact.brushing:
@@ -320,7 +324,7 @@ class BqplotCircleMode(BqplotSelectionTool):
                 # We use a tolerance of 1e-2 below to match the tolerance set in glue-core
                 # https://github.com/glue-viz/glue/blob/6b968b352bc5ad68b95ad5e3bb25550782a69ee8/glue/viewers/matplotlib/state.py#L198
                 if self._strict_circle:
-                    roi = CircularROI(xc=xc, yc=yc, radius=np.sqrt((rx**2 + ry**2) * 0.5))
+                    roi = TrueCircularROI(xc=xc, yc=yc, radius=np.sqrt((rx**2 + ry**2) * 0.5))
                 elif np.allclose(rx, ry, rtol=1e-2):
                     roi = CircularROI(xc=xc, yc=yc, radius=(rx + ry) * 0.5)
                 else:
@@ -332,6 +336,8 @@ class BqplotCircleMode(BqplotSelectionTool):
     def update_from_roi(self, roi):
         if isinstance(roi, CircularROI):
             rx = ry = roi.radius
+            if isinstance(roi, TrueCircularROI) and self._strict_circle is not False:
+                self._strict_circle = True
         elif isinstance(roi, EllipticalROI):
             if self._strict_circle:
                 rx, ry = np.sqrt((roi.radius_x ** 2 + roi.radius_y ** 2) * 0.5)
@@ -364,7 +370,7 @@ class BqplotTrueCircleMode(BqplotCircleMode):
 
         super().__init__(viewer, **kwargs)
 
-        if not kwargs.pop('strict_circle', True):
+        if kwargs.pop('strict_circle', True) is False:
             raise ValueError('BqplotTrueCircleMode cannot have strict_circle=False')
         self._strict_circle = True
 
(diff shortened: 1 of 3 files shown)


camipacifici:
Hi @dhomeier, I tried testing your solution here, but I am encountering an error when opening jdaviz in stand alone mode:
Traceback (most recent call last):
  File "/Users/cpacifici/opt/miniconda3/envs/prderek/bin/jdaviz", line 5, in <module>
    from jdaviz.cli import _main
  File "/Users/cpacifici/opt/miniconda3/envs/prderek/lib/python3.11/site-packages/jdaviz/__init__.py", line 16, in <module>
    from jdaviz.configs.specviz import Specviz  # noqa: F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cpacifici/opt/miniconda3/envs/prderek/lib/python3.11/site-packages/jdaviz/configs/__init__.py", line 1, in <module>
    from .cubeviz import *  # noqa
    ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cpacifici/opt/miniconda3/envs/prderek/lib/python3.11/site-packages/jdaviz/configs/cubeviz/__init__.py", line 1, in <module>
    from .plugins import *  # noqa
    ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cpacifici/opt/miniconda3/envs/prderek/lib/python3.11/site-packages/jdaviz/configs/cubeviz/plugins/__init__.py", line 1, in <module>
    from .tools import *  # noqa
    ^^^^^^^^^^^^^^^^^^^^
  File "/Users/cpacifici/opt/miniconda3/envs/prderek/lib/python3.11/site-packages/jdaviz/configs/cubeviz/plugins/tools.py", line 7, in <module>
    from jdaviz.configs.imviz.plugins.tools import _MatchedZoomMixin
  File "/Users/cpacifici/opt/miniconda3/envs/prderek/lib/python3.11/site-packages/jdaviz/configs/imviz/__init__.py", line 1, in <module>
    from .plugins import *  # noqa
    ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cpacifici/opt/miniconda3/envs/prderek/lib/python3.11/site-packages/jdaviz/configs/imviz/plugins/__init__.py", line 2, in <module>
    from .viewers import *  # noqa
    ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cpacifici/opt/miniconda3/envs/prderek/lib/python3.11/site-packages/jdaviz/configs/imviz/plugins/viewers.py", line 15, in <module>
    from jdaviz.configs.default.plugins.viewers import JdavizViewerMixin
  File "/Users/cpacifici/opt/miniconda3/envs/prderek/lib/python3.11/site-packages/jdaviz/configs/default/__init__.py", line 1, in <module>
    from .plugins import *  # noqa
    ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cpacifici/opt/miniconda3/envs/prderek/lib/python3.11/site-packages/jdaviz/configs/default/plugins/__init__.py", line 1, in <module>
    from .viewers import *  # noqa
    ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/cpacifici/opt/miniconda3/envs/prderek/lib/python3.11/site-packages/jdaviz/configs/default/plugins/viewers.py", line 5, in <module>
    from glue_jupyter.bqplot.scatter.layer_artist import BqplotScatterLayerState
ImportError: cannot import name 'BqplotScatterLayerState' from 'glue_jupyter.bqplot.scatter.layer_artist' (/Users/cpacifici/opt/miniconda3/envs/prderek/lib/python3.11/site-packages/glue_jupyter/bqplot/scatter/layer_artist.py)

I think I have the exact workflow where I see the problem. It happens when I pop out the viewer from the notebook. When popped out, the x,y aspect ratio is not retained (https://github.com/spacetelescope/jdaviz/issues/2187) and that is why a circular subset becomes an ellipse. We need to fix the aspect ratio bug, but I still think that a circle should be a circle regardless. Let me know if I need to do something different to test this PR. I can send a video of how I trigger the problem if that can be helpful. Thank you!

dhomeier:
> I think I have the exact workflow where I see the problem. It happens when I pop out the viewer from the notebook. When popped out, the x,y aspect ratio is not retained ([spacetelescope/jdaviz#2187](https://github.com/spacetelescope/jdaviz/issues/2187)) and that is why a circular subset becomes an ellipse. We need to fix the aspect ratio bug, but I still think that a circle should be a circle regardless. Let me know if I need to do something different to test this PR. I can send a video of how I trigger the problem if that can be helpful. Thank you!

Thanks for the details, I had been missing that you were creating a new window for the viewer here as well!
I can reproduce the `EllipticalROI` created originally, and could confirm that with this PR I get a `CircularROI` when using `'bqplot:truecircle'` in the viewer. This also remains when moving the created subset, but still changes to a `EllipticalROI` when I am resizing it later. This probably still needs fixing, have to check if it's an issue within the glue-jupyter tools as well.

Your failure in launching jdaviz in the first place is due to the missing fix for `BqplotScatterLayerState` which I pushed separately to #375, since it is a more immediate bugfix. You'll have to pull that in addition; if you have problems with that I might push the commit here as well while this is still in draft stage.

dhomeier:
Subset after move
Subset after resize

The first screenshot shows region properly returning the `CircularROI` after dragging and moving it, while in the second after resizing it has switched to `EllipticalROI`. My take is that the resize handle still creates a (new) plain `BqplotCircleMode`, but I could not find this defined anywhere in Jdaviz – is this a default operation from glue-jupyter?

pllim:
@dhomeier , this line needs to be changed in Imviz to use your new feature:

https://github.com/spacetelescope/jdaviz/blob/9f12e594d54a4fb3fdbcadf26687e34b4297a69c/jdaviz/configs/imviz/plugins/viewers.py#L27

@camipacifici , your error is from https://github.com/glue-viz/glue-jupyter/pull/363#pullrequestreview-1528650930 and looks like it is being fixed over at https://github.com/glue-viz/glue-jupyter/pull/375 .

dhomeier:
> @camipacifici , your error is from #363 (review) and looks like it is being fixed over at #375 .

Yes it should. I'll probably have the final version raise a `DeprecationWarning`, so this needs to be checked for your test suite – I was unable to run Jdaviz tests so far because I could not get it to ignore another Deprecation coming in from `echo`.

pllim:
> because I could not get it to ignore another Deprecation coming in from echo

To get past that, you can temporarily ignore that echo warning in Jdaviz `pyproject.toml` in the pytest filterwarnings section. We are not seeing this in CI, so maybe you can also try to upgrade echo or something.

dhomeier:
> To get past that, you can temporarily ignore that echo warning in Jdaviz `pyproject.toml` in the pytest filterwarnings section. We are not seeing this in CI, so maybe you can also try to upgrade echo or something.

Yes, putting it there rather than playing around in `setup.cfg` got around it (maybe the filterwarnings there should be removed?). AFAICS you are pinning to `echo==0.5`, but I am getting this and another `pkg_resources` one both with 0.5 and 0.8.0 – it probably should be fixed by glue-viz/echo#36, which is not merged yet.
But anyway after running next into the _Jupyter is migrating its paths to use standard platformdirs_ warning, my mambaforge  python segfaults on running the tests, so I'll leave it up to you to test the updated #375 (non-public, deprecated `BqplotScatterLayerState`) for now. Can maybe test with my work laptop tomorrow.

pllim:
Re: https://github.com/glue-viz/glue-jupyter/pull/376#issuecomment-1661079948

Are you sure you are not working off an old copy of Jdaviz code? We don't pin to `echo==0.5` on `main`. See https://github.com/spacetelescope/jdaviz/blob/c3168d4dd562ba143ca86ed41178177d685fccf7/pyproject.toml#L17

But the duplication in `setup.cfg` is unexpected and I fear it might be a botched rebase on very old PR that got merged. I will open issue about that on Jdaviz. Good catch!

dhomeier:
> Are you sure you are not working off an old copy of Jdaviz code? We don't pin to `echo==0.5` on `main`.

No, was checking what you were running in your CI, but misread `echo>=0.5.0` for `echo==0.5` – it was installing 0.8.0 indeed.
Should get that fix in for `echo` anyway.

pllim:
@dhomeier , I created https://github.com/spacetelescope/jdaviz/pull/2332 to test this downstream. Seems to work as expected. Thanks!

pllim (approved):

pllim:
@dhomeier , can you please rebase to pick up #375 here? Thanks!

dhomeier:
Already did; something must have stalled when pushing.

pllim:
The third commit after rebase looks unrelated.

dhomeier:
> The third commit after rebase looks unrelated.

Yes, but necessary for `BqplotScatterLayerState` not to raise a `UserWarning`, which would break Jdaviz tests.
Can put this in its own follow-up PR if this one takes too long.

dhomeier:
> Re: [#376 (comment)](https://github.com/glue-viz/glue-jupyter/pull/376#issuecomment-1642500383)
> 
> What does "resizing" mean? How and what do you "resize"?

In the image example the (only) Subset0, originally linked to a `CircularROI`. When placing the mouse over the perimeter instead of inside, so resize mode is activated instead of moving, once the radius is changed it becomes the `EllipticalROI` in the second image.

dhomeier:
Will rebase again once #379 is approved and merged.

pllim:
What...? I never realized you can resize it by clicking exactly on the edge! But I have a feeling it is the same code as moving, which is here:

https://github.com/glue-viz/glue-jupyter/blob/f0270ab31211184c07f410086b64830920b28eb5/glue_jupyter/bqplot/common/tools.py#L585-L587

pllim:
But I see that you already modified that part of the code, so I dunno. @astrofrog maybe can tell us.

pllim (changes requested):
I can reproduce the "ooops back to ellipse" bug now and I agree it needs to be fixed before we can merge this. Thanks!

dhomeier:
Yes; don't remember if I had an immediate reason to separate the cases or if it was just an attempt to make sure a circle always remains a circle everywhere.

dhomeier:
> What...? I never realized you can resize it by clicking exactly on the edge! But I have a feeling it is the same code as moving, which is here:

I think it must have an additional check somewhere, as `roi.contains(x, y)` would match anything _inside_ the ROI, while resizing evidently is triggered by something like `np.allclose((x-roi.xc)**2+(y-roi.yc)**2, roi.radius**2)` for points close to the edge – which would be different for other shapes of ROI.

astrofrog:
If I use https://github.com/spacetelescope/jdaviz/pull/2332 and this PR I can't manage to reproduce the fact it changes back to an ellipse when dragging or resizing the circle. It always stays as a circular ROI. Could one of you show me a screencast of the bug?

The feature of resizing the circle is built in to the bqplot elliptical selector, we get it 'for free' similar to moving the ROI.

glue_jupyter/bqplot/common/tools.py
```diff
@@ -582,7 +603,10 @@ def press(self, x, y):
             if layer.visible and isinstance(subset_state, RoiSubsetState):
                 roi = subset_state.roi
                 if roi.defined() and roi.contains(x, y):
-                    if isinstance(roi, (EllipticalROI, CircularROI)):
+                    if isinstance(roi, EllipticalROI):
+                        self._active_tool = BqplotEllipseMode(
+                            self.viewer, roi=roi, finalize_callback=self.release)
+                    elif isinstance(roi, CircularROI):
```
astrofrog, line 609: Is the issue related to the code here in that in the case where this was created from the true circle mode, we should actually be instantiating BqplotTrueCircleMode? I don't understand why I don't see the issue though.

astrofrog:
Ok I can reproduce the issue and indeed I think the issue is in my in-line comment above.

dhomeier:
So how to find out if the `CircularROI` originated from a `BqplotCircleMode` or `BqplotTrueCircleMode` (or access the parent's `_strict_circle` property) here?

pllim:
I am beginning to think we should get rid of this "secret" tool altogether to move/resize after creation. With "Subset Tools" plugin in Jdaviz now to enable more flexible editing, this tool is more of a burden and than gift now. Given that Tom and I were both unaware of the resize feature, probably no one is even using it. 

I have an idea downstream to have a new explicit tool for "click to move" that would utilize the common API in glue-core that would also allow moving composite subsets, which this tool cannot do.

astrofrog:
> So how to find out if the `CircularROI` originated from a `BqplotCircleMode` or `BqplotTrueCircleMode` (or access the parent's `_strict_circle` property) here?

Should we maybe add a private attribute on CircularROI to indicate whether it was created by the strict mode?

pllim:
FWIW found my old PR:

* https://github.com/spacetelescope/jdaviz/pull/2202

dhomeier:
> > Are you sure you are not working off an old copy of Jdaviz code? We don't pin to `echo==0.5` on `main`.
> 
> No, was checking what you were running in your CI, but misread `echo>=0.5.0` for `echo==0.5` – it was installing 0.8.0 indeed.

@pllim for the record, I am seeing this test failure on my Intel Mac:
```
ImportError while loading conftest '/Users/derek/git/jdaviz/conftest.py'.
conftest.py:1: in <module>
    from jdaviz import __version__
jdaviz/__init__.py:15: in <module>
    from jdaviz.app import *  # noqa: F401, F403
jdaviz/app.py:17: in <module>
    from echo import CallbackProperty, DictCallbackProperty, ListCallbackProperty
../../opt/mambaforge/envs/jdaviz/lib/python3.11/site-packages/echo/__init__.py:4: in <module>
    from .version import *  # noqa
../../opt/mambaforge/envs/jdaviz/lib/python3.11/site-packages/echo/version.py:1: in <module>
    from pkg_resources import get_distribution, DistributionNotFound
../../opt/mambaforge/envs/jdaviz/lib/python3.11/site-packages/pkg_resources/__init__.py:118: in <module>
    warnings.warn(
E   DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
```
but not on the M1 (with same filters), though with `DeprecationWarning` activated manually I get
```python
>>> from jdaviz import __version__
../../opt/mambaforge/envs/jdaviz/lib/python3.11/site-packages/echo/version.py:1: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
  from pkg_resources import get_distribution, DistributionNotFound
../../opt/mambaforge/envs/jdaviz/lib/python3.11/site-packages/pkg_resources/__init__.py:2871: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('mpl_toolkits')`.
Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
  declare_namespace(pkg)
```
both on the i5 and the M1; no idea why conftest.py is picking it up on the older, but not on the newer machine or CI.
Anyway the upstream fix should be in `echo` soon enough, though there are similar ones coming e.g. from `gwcs` next...

dhomeier:
Yes, did not want to follow up on that here, just leave a record of what may or may not be going on with the test suite (some other package already intercepting the warnings in some setups?)

dhomeier:
> I do not want to go to that Dependency Hell so early on a Monday.

Me neither ;-) Anyway this was actually just to make sure #379 did not break any tests; and I do have access to a working test env now or know how to create one.

dhomeier:
> Should we maybe add a private attribute on CircularROI to indicate whether it was created by the strict mode?

That would require syncing with that attribute to be added in `glue.core`, right? But I haven't been able to think of another solution; we only have the information available through the ROI or `RoiSubsetState` at this point, right? Or is there any way to retrieve the `tool_id`?

dhomeier:
> Should we maybe add a private attribute on CircularROI to indicate whether it was created by the strict mode?

I have now implemented your idea from the SME meeting defining this as a class, so that initialising from a `TrueCircularROI` will also set `self._strict_circle = True` by default (for some reason special-casing this in https://github.com/glue-viz/glue-jupyter/blob/f0270ab31211184c07f410086b64830920b28eb5/glue_jupyter/bqplot/common/tools.py#L585
breaks the drag-and-move mode when first initialising the brush selector, so I moved the functionality to `update_from_roi` instead).
Having `strict_circle` also as a kwarg complicates things a bit in this version, and has become pretty much redundant anyway, so perhaps it should just be removed?
I have verified resizing in the Voila app and ran Jdaviz tests locally, but did not check against the non-Imviz case as in
https://github.com/glue-viz/glue-jupyter/pull/370#issuecomment-1628892239

glue_jupyter/bqplot/common/tools.py
```diff
@@ -348,6 +360,21 @@ def activate(self):
...
+class BqplotTrueCircleMode(BqplotCircleMode):
+
+    tool_id = 'bqplot:truecircle'
+    tool_tip = 'Define a strictly circular region of interest'
+
+    def __init__(self, viewer, roi=None, finalize_callback=None, **kwargs):
+
+        super().__init__(viewer, **kwargs)
+
+        if kwargs.pop('strict_circle', True) is False:
```
pllim, line 373: Yeah, I agree this is redundant. What do you think, @astrofrog ?

astrofrog: Yes agree

dhomeier:
We don't have any tests directly calling the `Bqplot*Mode` interfaces directly, should we?
Otherwise if Jdaviz confirms this fixes the issue fully, it should be ready to go now.

pllim:
I did add a test here: https://github.com/glue-viz/glue-jupyter/pull/370

Maybe you can copy it over to suit your fix.

glue_jupyter/bqplot/tests/test_bqplot.py
```diff
@@ -5,6 +5,7 @@
 from nbconvert.preprocessors import ExecutePreprocessor
 from glue.core import Data
 from glue.core.roi import EllipticalROI
+from ..common.tools import TrueCircularROI
```
pllim, line 8: Do we still need relative import in Python 3?

pllim:
Oh, you mean test for mode, not the tool, nevermind, please ignore my comment about test. I don't need new tests beyond what you added here.

glue_jupyter/bqplot/tests/test_bqplot.py
dhomeier: Dunno, is it discouraged now?

dhomeier:
I did pull in your test and added another one for the TrueCircle version; should have listed you as co-author.

pllim:
Thanks, @dhomeier ! Please let us know when this feature is released, so I can revisit the downstream PR.

dhomeier:
@astrofrog how about a bugfix release once #384 and #385 are in, and perhaps see if we can get #383 finished within a day or so?

dhomeier:
Seemed it was considered a bug 😁

But yes, it is new functionality, and also a new tool in the standard notebook – maybe I should try to find an alternative icon for that.
Screenshot 2023-08-15 at 3 47 46 pm

pllim:
> although maybe a minor version rather than bugfix since this is a new feature?

Yeah, strictly speaking, it is a new feature though it acts like a bug in Jdaviz (mainly due to user confusion).

dhomeier:
@astrofrog, in view of the screenshot above (which is from #383) I am wondering which of the new subset tools should remain enabled by default. I had added them just because I did not know how to make them available in the tests otherwise, but I think the "circular annulus" would be genuinely useful. But "circular" and "strictly circular" may be more confusing in the present form; perhaps renaming it again to "persistent circular" would be a bit clearer, but I can't think of a good icon to symbolise that (maybe changing the non-persistent one to something combining circular and elliptic shapes). Or is there a way to add this to `tools` just for the tests?

astrofrog:
The annulus one would only make sense to include by default if we had a way to edit the inner radius - which I don't think we do here? (but there is in jdaviz)

You can add some of these just for tests by making a test viewer subclass where you edit the tools attribute - I think  that's probably the easiest way.

For now I don't think the strict circular mode should be enabled by default until we've had a bit more time to think about it - for instance it will only really be intuitive in the image viewer if aspect is equal, but if aspect is auto it will be a bit confusing. For now I would prefer to not expose this except in jdaviz.

dhomeier:
> The annulus one would only make sense to include by default if we had a way to edit the inner radius - which I don't think we do here? (but there is in jdaviz)

Isn't there a way to access the subset through `viewer.layers`? But that's certainly not a viable interface for the standard notebook setup.

> You can add some of these just for tests by making a test viewer subclass where you edit the tools attribute - I think that's probably the easiest way.

I've tried to do that, but could not figure out yet how the `viewer.toolbar.tools` dictionary is produced from the `tools` list in viewer.py, and correspondingly what's the correct way to modify it – I suppose this is done somewhere inside `bqplot`?

Carifio24:
@dhomeier If it helps, the `viewer.toolbar.tools` dictionary is created via the `add_tool` calls [here](https://github.com/glue-viz/glue-jupyter/blob/main/glue_jupyter/view.py#L148). So for a subclass all you should need to is define the `tools = ["some_tool_id_1", "some_tool_id_2", ...]` attribute on the viewer class and the toolbar will get populated with the corresponding tools (plus base class tools, unless you set `inherit_tools = False`).

dhomeier:
Thanks @Carifio24 , exactly the bit of info I needed! In ca50995 tests are working with just calling `add_tool` on the viewer instance, so creating a new subclass should not even be required.

PR #378, opened 2023-08-01
Restore BqplotScatterLayerState so old import does not break downstream

pllim:
Restore BqplotScatterLayerState so old import does not break downstream. I find this the least intrusive way to fix the error in Jdaviz but is it also too hacky?

Close #375 

cc @dhomeier @astrofrog


dhomeier:
It looks more elegant in a way, though indeed a bit hackish. But we agreed in #375 to deprecate its use, I don't see how to easily implement that in this approach. I accidentally raised a `UserWarning` over there, so that still needs a follow-up – with a `DeprecationWarning` it does not seem to affect Jdaviz tests, as the import is probably not directly tested there.

pllim:
Jdaviz cannot use other fixes in dev until this error goes away, so in a way, this is 🔥 . If you cannot deprecate it easily and need more time, should we just merge this first so Jdaviz can move on for now?

pllim:
Ah, wait, but you already merged #375 . Nvm then.

PR #383, opened 2023-08-12
Prevent destruction of `CircularAnnulusROI` by resizing below `inner_radius`

dhomeier:
## Description
https://github.com/glue-viz/glue-jupyter/pull/376#issuecomment-1664217504 brought to light an issue with the little-documented **resize** functionality of the bqplot brush tool when modifying `BqplotCircularAnnulusMode`: since the update from the modified `self._roi` only changes the outer and not the inner radius, it is possible to decrease the former below an invalid value `outer_radius <= inner_radius`, as a result destroying the subregion. This PR prevents this by keeping `inner_radius` just below `outer_radius`.
Such a minimal-width circular annulus will typically become impossible to manipulate interactively, but it still exists and can e.g. be restored to some sensible inner radius via the menu.

Screenshot 2023-08-12 at 4 54 29 pm

Screenshot 2023-08-12 at 5 01 40 pm

I have added a basic test for creating this subset type, but was unable to emulate even the `move` operation this way, let alone resizing – pointers how to do this with `tool.interact` welcome.

### DIFF
--- a/glue_jupyter/bqplot/tests/test_bqplot.py
+++ b/glue_jupyter/bqplot/tests/test_bqplot.py
@@ -4,7 +4,7 @@ import numpy as np
 from numpy.testing import assert_allclose
 from nbconvert.preprocessors import ExecutePreprocessor
 from glue.core import Data
-from glue.core.roi import EllipticalROI
+from glue.core.roi import CircularAnnulusROI, EllipticalROI
 from ..common.tools import TrueCircularROI
 
 DATA = os.path.join(os.path.dirname(__file__), 'data')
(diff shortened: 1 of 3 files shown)


glue_jupyter/bqplot/tests/test_bqplot.py
```diff
@@ -305,6 +305,33 @@ def test_imshow_elliptical_brush(app, data_image):
...
+    assert_allclose(roi.xc, 151.00)
+    assert_allclose(roi.yc, 276.75)
+    assert_allclose(roi.outer_radius, 211.375)
+    assert_allclose(roi.inner_radius, 105.6875)
+
+    tool.interact.brushing = True
+    tool.interact.selected = [(150.0, 275.0), (150.0, 325.0)]
+    tool.interact.brushing = False
+    assert_allclose(roi.xc, 151.00)
+    # assert_allclose(roi.yc, 326.75)  # trying to use `move`, but this is not how it works...
```
pllim, line 330: I don't understand this comment. Is there a bug that need fixing here?

dhomeier: The "bug" is me not knowing how to work with `tool.interact` to reproduce the interaction of clicking inside the subset, then dragging it 50 pixels upwards and release it (i.e. do a `move` of +50 pixels in y). Just left here as a reminder, but if I don't find a solution, the commented part can be removed again.

astrofrog (changes requested):
Seems reasonable to me, thanks! But as noted in https://github.com/glue-viz/glue-jupyter/pull/376#issuecomment-1682074494 I don't think we should be adding more default tools for now.

pllim:
Is there a way to test stuff without having to add things to default tools proper?

astrofrog (approved):

PR #391, opened 2023-08-18
Complement BqplotLassoMode with genuine PolygonMode and enable dragging with BrushSelector

dhomeier:
## Description
The present behaviour of the new `BqplotPolygonMode` still appears somewhat buggy to me: in
https://github.com/glue-viz/glue-jupyter/blob/3b264c76b1dd4ff8004646d8beac15994a4da459/glue_jupyter/bqplot/common/tools.py#L211-L215
the intention seems to be to not modify the existing subset at all when selected, but since `ROIClickAndDrag` is handing both rectangular and polygonal ROIs to `BqplotRectangleMode`, the latter will just replace the polygon with a rectangle from the bounding box and then let that be moved around (at least in jupyterlab; in the regular notebook the `BqplotRectangleMode` still seems to react very inconsistently, as indicated in
https://github.com/glue-viz/glue-jupyter/blob/3b264c76b1dd4ff8004646d8beac15994a4da459/glue_jupyter/bqplot/common/tools.py#L169-L171
but that is a separate issue).

Passing the polygon back to `BqplotPolygonMode` instead requires the `interact` property to be set at a minimum, so at that point I made an attempt to actually implement the `move` functionality – the comment above stated it might not make much sense, but I don't see much less use in it than dragging an ellipse or circle.
At the moment, for lack of a better option, this is using a standard (rectangular) `BrushSelector` to set a target point at the centre of the rectangle and then shift the polygon's centroid to that point:
Screenshot 2023-08-18 at 8 24 10 pm
Screenshot 2023-08-18 at 8 25 01 pm
Screenshot 2023-08-18 at 8 25 15 pm

There might certainly be intuitively better tools for this, to be discussed – pinging @jfoster17.


dhomeier:
Somewhat related, since the existing mode really is rather a lasso mode (continuously adding points), I've been thinking about also providing a discrete polygon mode that allows adding distinct points by clicking. This can be done based on the same class, so we might have a `BqplotLassoMode` with `tool_id = 'bqplot:lasso'` for the current behaviour and `BqplotPolygonMode` with `tool_id = 'bqplot:polygon'` for easily creating triangles, pentagons etc.

dhomeier:
I've added this as a proof of concept in the second commit for now.

Carifio24:
The ROI moving is working well for me in the image viewer, but it's not possible to move it in the scatter viewer as I couldn't find a way to activate the `BrushSelector`.

I like the discrete polygon mode - it's a nice option to have in addition to the lasso drag mode. My one comment is that it would be nice to have another way (maybe something like pressing enter?) to close the polygon rather than needing to return to the starting vertex (sometimes I forgot where I had started 😅). Also, I would guess that we want to re-add the lasso mode to the scatter viewer as well.

dhomeier:
> The ROI moving is working well for me in the image viewer, but it's not possible to move it in the scatter viewer as I couldn't find a way to activate the `BrushSelector`. 

Me neither – I did not try before, but I could not find a way to activate it for rectangle or circular ROIs either as the region itself or its extent is simply not highlighted, so I suppose there is no equivalent to the click-inside-region to activate `ROIClickAndDrag`. Don't know if there is a straightforward way to add this for scatter mode (for all ROI types), but that's definitely a separate issue.

> I like the discrete polygon mode - it's a nice option to have in addition to the lasso drag mode. My one comment is that it would be nice to have another way (maybe something like pressing enter?) to close the polygon rather than needing to return to the starting vertex (sometimes I forgot where I had started 😅). 

The existing vertices should be displayed in real-time 😉 but I found it problematic myself having to hit the first node close enough, so it will already close the polygon once you deselect the button in the toolbar. Having a key for the same functionality would be convenient, but I'd have to investigate how to pass that to the ROI.

> Also, I would guess that we want to re-add the lasso mode to the scatter viewer as well.

Yes, that was just a side-effect of making the original `bqplot:polygon` the new polygon mode, but I think scatter should probably offer the same set of tools as image (or a superset, `[xy]range` may not make much sense in the image viewer).

dhomeier:
@astrofrog I am not sure what the function of the corresponding classes in `ipyvolume` is, as `IpyvolumeLassoMode` predates the introduction of the Bqplot tool quite a while, but I've found it safer to rewrite both Lasso and Polygon classes analogously to the `bqplot` ones.

PR #394, opened 2023-08-25
Make bqplot linear scatter sizes more closely match matplotlib viewer

### LINKED ISSUE #392: Disparity in linear scatter size between matplotlib and bqplot viewers
**Describe the bug**
The displayed sizes when using linear size mode are significantly different between the matplotlib (Qt) and bqplot viewers. In particular, I think that the matplotlib layer artist accounts for the fact that we are setting area sizes, while the bqplot layer artist doesn't.

**To Reproduce**
This could be reproduced with any data, but here are quick snippets:

In the Qt terminal:
```
from glue.core import Data
from glue_qt.viewers.scatter import ScatterViewer

x = [i for i in range(100)]
y = [i + 200 for i in range(100)]
data = Data(label='data', x=x, y=y)
dc.append(data)

viewer = application.new_data_viewer(ScatterViewer, data=data)
layer = viewer.layers[0]
layer.state.size_att = data.id['x']
layer.state.size_mode = 'Linear'
layer.state.size_vmin = 0
layer.state.size_vmax = 100
layer.state.color = '#0055ff'
layer.state.size_scaling = 2
```

In a Jupyter notebook:
```
from glue.core import Data
from glue_jupyter import jglue
from glue_jupyter.bqplot.scatter import BqplotScatterView

app = jglue()
session = app.session
dc = app.session.data_collection

x = [i for i in range(100)]
y = [i + 200 for i in range(100)]
data = Data(label='data', x=x, y=y)
dc.append(data)

viewer = app.new_data_viewer(BqplotScatterView, data=data)
layer = viewer.layers[0]
layer.state.size_att = data.id['x']
layer.state.size_mode = 'Linear'
layer.state.size_vmin = 0
layer.state.size_vmax = 100
layer.state.color = '#0055ff'
layer.state.size_scaling = 2
```

**Expected behavior**
The two plots should look basically the same. Obviously they won't look identical as we're using different frontends, but I would expect to see the same general behavior

**Screenshots**
Here's what this looks like in Qt glue:
SizeScalingMpl

and the same thing (equivalent state values) in the bqplot viewer:
SizeScalingBqplot

If I manually square each element of the scatter mark's size array, I get this, which looks quite similar to the matplotlib viewer:
SizeScalingBqplot_modified

**Details:**

- Operating System: Ubuntu 20.04
- Python version: 3.11.4
- Glue version: 1.13.1
- How you installed glue: pip

**Additional context**
As I mentioned above, I believe the disparity is due to the fact that the bqplot layer artist is missing an equivalent to [this line](https://github.com/glue-viz/glue/blob/main/glue/viewers/scatter/layer_artist.py#L437), as bqplot marks are also setting the area.

Carifio24:
This PR aims to resolve #392 by making the linear size scaling behave more similarly to that of the matplotlib viewer. To do this, I ended up not using the size scales for the layer, but rather following the matplotlib viewer's example and capping the size.

FWIW, my initial thought was to do something like
```
s = ensure_numerical(self.layer[self.state.size_att].ravel())
s = (s - self.state.size_vmin) / (self.state.size_vmax - self.state.size_vmin)
np.clip(s, self.state.size_vmin, self.state.size_vmax, out=s)
s *= 0.95
s += 0.05 * (self.state.size_vmax - self.state.size_vmin)
s *= self.scatter_mark.default_size
self.scatter_mark.size = s ** 2

self.scale_size_scatter.min = float_or_none(self.state.size_vmin ** 2)
self.scale_size_scatter.max = float_or_none(self.state.size_vmax ** 2)
```

but that gave results that differed from the matplotlib implementation when `size_vmin` was nonzero (like e.g. the example data in #392).

### DIFF
--- a/glue_jupyter/bqplot/scatter/layer_artist.py
+++ b/glue_jupyter/bqplot/scatter/layer_artist.py
@@ -91,11 +91,9 @@ class BqplotScatterLayerArtist(LayerArtist):
 
         # Scatter points
 
-        self.scale_size_scatter = bqplot.LinearScale()
         self.scale_color_scatter = bqplot.ColorScale()
         self.scales_scatter = dict(
             self.view.scales,
-            size=self.scale_size_scatter,
             color=self.scale_color_scatter,
         )
 
@@ -270,15 +268,15 @@ class BqplotScatterLayerArtist(LayerArtist):
                             self.state.size * self.state.size_scaling,
                         )
                         self.scatter_mark.size = None
-                        self.scale_size_scatter.min = 0
-                        self.scale_size_scatter.max = 1
                     else:
-                        self.scatter_mark.default_size = int(self.state.size_scaling * 25)
-                        self.scatter_mark.size = ensure_numerical(
-                            self.layer[self.state.size_att].ravel()
-                        )
-                        self.scale_size_scatter.min = float_or_none(self.state.size_vmin)
-                        self.scale_size_scatter.max = float_or_none(self.state.size_vmax)
+                        self.scatter_mark.default_size = int(self.state.size_scaling * 7)
+                        s = ensure_numerical(self.layer[self.state.size_att].ravel())
+                        s = (s - self.state.size_vmin) / (self.state.size_vmax - self.state.size_vmin)
+                        np.clip(s, 0, 1, out=s)
+                        s *= 0.95
+                        s += 0.05
+                        s *= self.scatter_mark.default_size
+                        self.scatter_mark.size = s ** 2
 
         if (
             self.state.vector_visible
--- a/glue_jupyter/bqplot/scatter/layer_artist.py
+++ b/glue_jupyter/bqplot/scatter/layer_artist.py
@@ -271,7 +271,8 @@ class BqplotScatterLayerArtist(LayerArtist):
                     else:
                         self.scatter_mark.default_size = int(self.state.size_scaling * 7)
                         s = ensure_numerical(self.layer[self.state.size_att].ravel())
-                        s = (s - self.state.size_vmin) / (self.state.size_vmax - self.state.size_vmin)
+                        s = ((s - self.state.size_vmin) /
+                             (self.state.size_vmax - self.state.size_vmin))
                         np.clip(s, 0, 1, out=s)
                         s *= 0.95
                         s += 0.05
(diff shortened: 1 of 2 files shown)


dhomeier:
I suppose the image comparison failure is not unexpected, as the produced figure should look different now – need to update the reference image then.

glue_jupyter/bqplot/scatter/layer_artist.py
```diff
@@ -270,15 +268,16 @@ def _update_visual_attributes(self, changed, force=False):
...
-                        )
-                        self.scale_size_scatter.min = float_or_none(self.state.size_vmin)
-                        self.scale_size_scatter.max = float_or_none(self.state.size_vmax)
+                        self.scatter_mark.default_size = int(self.state.size_scaling * 7)
+                        s = ensure_numerical(self.layer[self.state.size_att].ravel())
+                        s = ((s - self.state.size_vmin) /
+                             (self.state.size_vmax - self.state.size_vmin))
+                        np.clip(s, 0, 1, out=s)
+                        s *= 0.95
+                        s += 0.05
```
dhomeier, line 278: Those numbers (`size_scaling * 7`, `s` ranging from 0.05 to 1.0) are taken from comparison with the matplotlib defaults?
I was wondering if the division by 0 in

> scatter_density_mark.py:210: RuntimeWarning: invalid value encountered in divide
    normalized_counts = (self._counts - vmin) / (vmax - vmin)

was related, but that warning already pops up prior to this PR – I suppose that's bad user input really.

Carifio24: Yeah, the 7 is a magic constant that I thought gave the closest match to the matplotlib layer artist - basically it's meant to serve a role similar what 30 does [here](https://github.com/glue-viz/glue/blob/main/glue/viewers/scatter/layer_artist.py#L428).

Carifio24:
I'm not super familiar with the CircleCI system yet, but I think you're right that we just need to update the hashes. Once we're satisfied with all the code changes I'll take care of that.

glue_jupyter/bqplot/scatter/layer_artist.py
dhomeier: If the parent `LayerArtist` is crafted that way, matching that makes sense. I'm good with the implantation so far, let's wait a little if @astrofrog has any comments.

dhomeier:
Go ahead with updating the hashes if you wish; once the CI is fixed this should be good to go in for the 0.18.0 release.

dhomeier:
> I'm not super familiar with the CircleCI system yet, but I think you're right that we just need to update the hashes. Once we're satisfied with all the code changes I'll take care of that.

Thanks! Struggling to make sense of it myself – image comparison is now passing with the expected hash, though the figure_report still shows a different (the original) baseline image. From the pytest invocation I take that the reference figures are pulled from https://github.com/glue-viz/glue-jupyter-visual-tests; this in turn seems to be automatically updated when a change in `py311-test-visual.json` is pushed to `main`, so I'd just merge and see if this works.

I was more worried about the docs build still failing, but seeing now that readthedocs tries to use an obsolete .readthedocs.yml with Python 3.6. Should have rebased to pull in #386, but I think this will also sort itself out on merging.

PR #396, opened 2023-08-28
Preserve rotation angle under dragging for ellipse and rectangle selection tools

pllim:
In Jdaviz, you can rotate ellipse and rectangle ROIs using Subset Tools plugin. After that is done, and user wants to click and drag the shape, without this patch, the rotation resets. This patch preserves any existing rotation. Since I really don't want to implement theta support in JS (e.g., https://github.com/glue-viz/bqplot-image-gl/blob/main/js/lib/BrushEllipseSelector.js), during the dragging phase, the gray shape has no rotation, but the rotation comes back when you finalize the drag.

cc @astrofrog and @dhomeier 

[🐱](https://jira.stsci.edu/browse/JDAT-3717)

### DIFF
--- a/glue_jupyter/bqplot/common/tools.py
+++ b/glue_jupyter/bqplot/common/tools.py
@@ -136,6 +136,7 @@ class BqplotRectangleMode(BqplotSelectionTool):
                                       y_scale=self.viewer.scale_y,
                                       color=INTERACT_COLOR)
 
+        self._roi = kwargs.get("roi", None)
         if roi is not None:
             self.update_from_roi(roi)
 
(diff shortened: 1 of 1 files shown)


glue_jupyter/bqplot/common/tools.py
```diff
@@ -166,9 +170,6 @@ def update_from_roi(self, roi):
                 self.interact.selected_y = [np.min(roi.vy), np.max(roi.vy)]
             else:
                 raise TypeError(f'Cannot initialize a BqplotRectangleMode from a {type(roi)}')
-            # FIXME: the brush selector does not actually update unless the
-            # widget is resized/refreshed, see
-            # https://github.com/bloomberg/bqplot/issues/1067
```
dhomeier, line 171: Not sure if this can go – I am still, though inconsistently, seeing the rectangle position not updating properly (right now of all things it does not seem to move at all in the Imviz viewer!). No idea if this is related to the original issue, but I'd rather leave the reminder in for now.

dhomeier (changes requested):
Too bad again that moving cannot be tested currently, but otherwise this looks good – as mentioned, I am unable to check if this works for Rectangle, but for Ellipse it does as expected, so I'll trust that the former is solved for you as well.

glue_jupyter/bqplot/common/tools.py
pllim: Weird, it works for me. Huh. I'll add it back.

pllim:
Try test it over in Imviz. Install this branch, and then install Jdaviz, and test it there.

dhomeier:
> Try test it over in Imviz. Install this branch, and then install Jdaviz, and test it there.

That's what I did, or do you mean installing in different order makes any difference?
Seems totally non-deterministic anyway – at first moving did not update properly in the regular notebook, but worked in Jupyterlab; now it's functioning normally in the standard notebook, but there is no way to access `roi.rotate_by`, while in Imviz the rectangles (and only rectangles) can't be moved at all...

pllim:
I think as long as you can prove to yourself using ellipse, it is good enough? The logic is the same in both. If you want to, we can meet at some point and I can screenshare, but not today.

dhomeier (approved):
OK, I could confirm it in the standalone app, and then after relaunching everything, moving worked in the notebook as well!

pllim:
I created an issue upstream for the confusing "shadow." FYI. https://github.com/glue-viz/bqplot-image-gl/issues/106

PR #399, opened 2023-09-12
Allow native aspect ratio in ipyvolume viewers

Carifio24:
Something that was requested during the last glue meeting was to expose an option to use the native aspect ratio in the ipyvolume viewers. This PR does that.

The only subtlety here is that we have to calculate the aspect in terms of x, z, y coordinates to account for the order in which we're using the ipyvolume axes in glue.


astrofrog (approved):
Nice, thanks!

PR #403, opened 2023-10-14
EXP: Add colorbar to image maybe

pllim:
We are trying to add a colorbar for Imviz somewhere. The bqplot examples does not show that is it possible to render the colorbar separately. I tried to attach one to axes like the examples but this one does not crash but also does not show colorbar in Imviz. Need advice, @astrofrog, @maartenbreddels, or @mariobuikhuizen . 🙏


pllim:
Thanks! But from the discussions this morning, looks like we do not want to use ColorScale at all, so closing.

PR #404, opened 2023-10-23
Added regression test for bug that caused image viewer to turn green when an incompatible subset was present

astrofrog:
Regression test for bug fixed by https://github.com/glue-viz/glue/pull/2425 - at the moment the regression test should fail but I will do a new release of glue-core and update the minimum version here and make sure the test passes.


astrofrog:
Well I never got this to work, closing as not worth putting more effort into

PR #402, opened 2023-10-10
add histogram, scatter, and volume viewers to registry

kecnry:
This PR adds missing viewers (histogram, profile, volume, _and image_) to the viewer registry.

Issue #388, opened 2023-08-17
Properly populate viewer registry


astrofrog:
The viewer registry defined in this package is empty by default and only gets populated when viewers are imported. We should instead populate this registry via entry point plugins like we do for the Qt version of glue, and we need to make sure that e.g. the glue-wwt viewer gets included if installed.

astrofrog:
This is related to https://github.com/QuantStack/glue-jupyterlab/issues/121

PR #411, opened 2023-11-29
Add entry points for viewer registry

jfoster17:
# Populate viewer registry for all viewers

## Description

This uses entry points to register all the viewers for glue-jupyter into viewer_registry. This addresses #388 and makes all viewers accessible through the `new_data_viewer()` command in glue-jupyter.  I think this should be compatible with #402, because the viewer registry won't double-populate the same viewers. I have added `IpyvolumeScatterView` as `scatter3d`.

Adding an entry point for glue-wwt will be a separate PR in that repository. 

Not 100% about the names here. 'image' is a convenient name for calling in code, but is less descriptive than something like '2d image'. This is particularly relevant for 'scatter' (unqualified) versus 'scatter3d'.

### DIFF
--- a/glue_jupyter/bqplot/histogram/__init__.py
+++ b/glue_jupyter/bqplot/histogram/__init__.py
@@ -1,2 +1,6 @@
 from .layer_artist import *  # noqa
 from .viewer import *  # noqa
+
+
+def setup():
+    from viewer import BqplotHistogramView # noqa
--- a/glue_jupyter/ipyvolume/scatter/viewer.py
+++ b/glue_jupyter/ipyvolume/scatter/viewer.py
@@ -1,4 +1,6 @@
 from glue_jupyter.common.state3d import Scatter3DViewerState
+from glue_jupyter.registries import viewer_registry
+
 from .layer_artist import IpyvolumeScatterLayerArtist
 from .layer_style_widget import Scatter3DLayerStateWidget
 from ..common.viewer_options_widget import Viewer3DStateWidget
@@ -7,6 +9,7 @@ from ..common.viewer import IpyvolumeBaseView
 __all__ = ['IpyvolumeScatterView']
 
 
+@viewer_registry("scatter3d")
 class IpyvolumeScatterView(IpyvolumeBaseView):
 
     allow_duplicate_data = False
--- a/setup.cfg
+++ b/setup.cfg
@@ -28,6 +28,16 @@ install_requires =
     bqplot-gl
     scikit-image
 
+[options.entry_points]
+glue.plugins =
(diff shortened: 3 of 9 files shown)


glue_jupyter/bqplot/histogram/__init__.py
```diff
@@ -1,2 +1,6 @@
 from .layer_artist import *  # noqa
 from .viewer import *  # noqa
+
+
+def setup():
+    from viewer import BqplotHistogramView # noqa
```
maartenbreddels, line 6: I don't think this is needed, since * is already imported

maartenbreddels (commented):
Nice, i was missing this in glue-solara!

glue_jupyter/bqplot/histogram/__init__.py
jfoster17: Thanks. I think I got confused because I forgot that you have to manually rerun pip install to rebuild the entry points when I was developing this. And yes, this means we can simplify glue-solara to just build a list of viewers from the registry.

glue_jupyter/bqplot/histogram/__init__.py
```diff
@@ -1,2 +1,6 @@
 from .layer_artist import *  # noqa
 from .viewer import *  # noqa
+
+
+def setup():
+    pass
```
astrofrog, line 6: Ideally the ``setup`` function should actually be adding the viewer to the registry (rather than using the decorator) - @CyclingNinja is working on improvements to ``load_plugins`` that means that users will be able to opt-in to specific plugins, so we need the viewer to not be added if the entry point is not called.

setup.cfg
```diff
@@ -28,6 +28,16 @@ install_requires =
     bqplot-gl
     scikit-image
 
+[options.entry_points]
+glue.plugins =
+    histogram = glue_jupyter.bqplot.histogram:setup
+    image = glue_jupyter.bqplot.image:setup
+    profile = glue_jupyter.bqplot.profile:setup
+    scatter = glue_jupyter.bqplot.scatter:setup
```
astrofrog, line 36: These plugin names might conflict with the ones from glue-core - the entry point names should be unique across the glue ecosystem.

glue_jupyter/ipyvolume/scatter/viewer.py
```diff
@@ -7,6 +9,7 @@
 __all__ = ['IpyvolumeScatterView']
 
 
+@viewer_registry("scatter3d")
```
astrofrog, line 12: The registration of the viewer should be done in ``setup`` as mentioned above.

astrofrog (changes requested):
Good idea to define entry points but some comments below on implementation.

PR #412, opened 2023-11-30
Fix support for bitmap_visible in image viewer

astrofrog:
Fix the case where ``visible`` is ``True`` but ``bitmap_visible`` is ``False`` in image viewer


kecnry (approved):
Seems to do the trick - thanks!

dhomeier:
circleci failure is unrelated

dhomeier (approved):
CircleCI failure has shown up before; `publish` failure due to #413.

PR #415, opened 2023-12-05
Fix visual tests with solara

astrofrog:
I think this should get things running again


astrofrog:
Ok the visual tests are working though they are failing due to real issues. Let's see if we can get them passing in a separate PR.

PR #417, opened 2023-12-28
Support setting step value in glue-float-field

rosteen:
Currently `glue-float-field` doesn't pass `step` along to `text-field`, this PR fixes that.


mariobuikhuizen (approved):

dhomeier (approved):
Not familiar with this, but looks clear and straightforward enough!

Issue #418, opened 2024-01-02
Let spatial shape tool block resizing attempt


pllim:
This is a follow-up of this rejected PR:

* https://github.com/glue-viz/glue-jupyter/pull/416

Desired outcome:

* Let us tell the tool when it is okay to resize and when it is not to prevent spatial subset from disappearing (i.e., becomes very small suddenly) when user drags to move it around in Jdaviz.
* Or better yet, prevent wonky `selected_x` and `selected_y` inputs in the first place, though no one seems to know how to fix that.

dhomeier:
Since I cannot reproduce the phenomenon on any of my machines, feel tree to try the modification in #419

PR #419, opened 2024-01-03
Set Bqplot*Modes to observe only one `selected` state in BrushSelector

dhomeier:
## Description

Pushing this as a test to avoid possible race conditions in observing both `selected_x` and `selected_y` in the BrushSelector interaction that have been speculated to cause unwanted size changes of the selection in #418.

Since the reported bug is non-deterministic and I have not been able to reproduce it at all, I have no idea if this change will help, just keeping it here for further investigation.

### DIFF
--- a/glue_jupyter/bqplot/common/tools.py
+++ b/glue_jupyter/bqplot/common/tools.py
@@ -141,8 +141,7 @@ class BqplotRectangleMode(BqplotSelectionTool):
             self.update_from_roi(roi)
 
         self.interact.observe(self.update_selection, "brushing")
-        self.interact.observe(self.on_selection_change, "selected_x")
-        self.interact.observe(self.on_selection_change, "selected_y")
+        self.interact.observe(self.on_selection_change, "selected")
         self.finalize_callback = finalize_callback
 
     def update_selection(self, *args):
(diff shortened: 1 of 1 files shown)


pllim:
Won't all the calls to `interact.selected_x` and `interact.selected_y` also need to be changed?

dhomeier:
`selected` includes both, so any change to an element of either one should alert the `observe`, but for working with the coordinates using `selected_[xy]` should just function the same – at least it works as far as I could see.
https://github.com/bqplot/bqplot/blob/c186fcdbf3c5455c91cb750ab5a3f94accefda81/bqplot/interacts.py#L365-L369

pllim:
I won't have time for to test this week. Try Cubeviz example notebook if you cannot wait. I could reproduce it there but not consistently.

dhomeier:
Tried on 3 different macOS and Python versions now, with notebook, jupyterlab and standalone for Cubeviz and the original Imviz example, and never saw as much as a flicker in size – will have to wait then.

pllim:
@cshanahan1 , since you most recently ran into the "disappearing subset" problem, do you want to see if this PR helps? Given how non-deterministic it is, better have multiple devs confirm.

pllim:
Maaaybe this fixes some of the problem, but it does not prevent the case where user is dragging something around very fast with repeated down/up clicks. The cursor can accidentally be right at the edge, where the "resize" mode would activate. Then when user intends to move, the app thinks user is resizing, resulting in the shape getting very big or small.

pllim:
Despite https://github.com/glue-viz/glue-jupyter/pull/419#issuecomment-1909042193 , listening to just one thing is better than having race condition, right? Should you take this out of draft?

dhomeier:
Cursor being too close to the edge so it does switch into resize mode seems out of scope for the race condition issue.
But I am not sure I can follow re. the rapid up/down clicks – maybe I need to test this with an actual mouse; with the trackpad there is always some pause before I can lock into drag mode with a new tap.
Anyway if there is at least some evidence this can help, I am changing status now.

astrofrog (approved):
Seems sensible, thanks!

PR #420, opened 2024-01-31
Set `InteractCheckableTool._roi` to `None` on `deactivate`

dhomeier:
## Description
Reset any `_roi` in an `InteractCheckableTool` when calling `deactivate`.
This is to avoid inheriting changed attributes like rotation angle from previously created selections. It will not reset the attributes when creating new ROIs with the same (left open) tool instance.

### DIFF
--- a/glue_jupyter/bqplot/common/tools.py
+++ b/glue_jupyter/bqplot/common/tools.py
@@ -48,6 +48,8 @@ class InteractCheckableTool(CheckableTool):
 
     def deactivate(self):
         self.viewer._mouse_interact.next = None
+        if hasattr(self, '_roi'):
+            self._roi = None
 
 
 class BqplotSelectionTool(InteractCheckableTool):


pllim:
Is this trying to solve the same problem that I approached differently in https://github.com/glue-viz/glue-jupyter/pull/416 ? Or is this for something else?

rosteen:
> Is this trying to solve the same problem that I approached differently in #416 ? Or is this for something else?

This is trying to solve the problem I was hacking around in https://github.com/spacetelescope/jdaviz/pull/2639.

rosteen:
@dhomeier This looks good to go in to me, I confirmed on the Jdaviz side that it fixes the unwanted behavior.

pllim:
tl;dr -- This does not reintroduce the problem where the angle resets when you click and drag a rotated shape, does it?

rosteen:
> tl;dr -- This does not reintroduce the problem where the angle resets when you click and drag a rotated shape, does it?

The existing problem of the gray preview not matching the rotation of the actual subset persists, but it's not worse than the current state. The dragged subset retains its angle in my testing.

astrofrog (approved):
Looks good, thanks!

Issue #421, opened 2024-02-13
ImportError: cannot import name 'get_cmap' from 'matplotlib.cm'


pllim:
Seen with `matplotlib==3.9.0.dev0`. Example log: https://github.com/spacetelescope/jdaviz/actions/runs/7892787965/job/21540037181?pr=2703

```
glue_jupyter/state_traitlets_helpers.py:9: in <module>
    from matplotlib.cm import get_cmap
E   ImportError: cannot import name 'get_cmap' from 'matplotlib.cm' (matplotlib/cm.py)
```

https://github.com/glue-viz/glue-jupyter/blob/ed24ff3999567a5dfab58615822adc56c6a12084/glue_jupyter/state_traitlets_helpers.py#L9

See https://github.com/matplotlib/matplotlib/blob/main/doc/api/next_api_changes/removals/26965-ER.rst

PR #422, opened 2024-02-13
Use ``mpl.colormaps.get_cmap`` for compatibility with matplotlib 3.9

pllim:
Fix https://github.com/glue-viz/glue-jupyter/issues/421 , maybe.

### DIFF
--- a/glue_jupyter/state_traitlets_helpers.py
+++ b/glue_jupyter/state_traitlets_helpers.py
@@ -1,12 +1,16 @@
 import json
-import matplotlib
 import traitlets
 from collections import defaultdict
 from traitlets.utils.bunch import Bunch
 from glue.core.state_objects import State
 from glue.core import BaseCartesianData, Subset, ComponentID
+from glue.utils.matplotlib import MATPLOTLIB_GE_36
 from echo import CallbackList, CallbackDict
 from matplotlib.colors import Colormap
+if MATPLOTLIB_GE_36:
+    from matplotlib import colormaps
+else:
+    from matplotlib import cm as colormaps
 
 MAGIC_IGNORE = '611cfa3b-ebb5-42d2-b5c7-ba9bce8b51a4'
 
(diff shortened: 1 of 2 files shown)


glue_jupyter/state_traitlets_helpers.py
```diff
@@ -1,12 +1,12 @@
 import json
+import matplotlib
 import traitlets
 from collections import defaultdict
 from traitlets.utils.bunch import Bunch
 from glue.core.state_objects import State
 from glue.core import BaseCartesianData, Subset, ComponentID
 from echo import CallbackList, CallbackDict
 from matplotlib.colors import Colormap
```
dhomeier, line 9:
```suggestion
import traitlets
from collections import defaultdict
from traitlets.utils.bunch import Bunch
from glue.core.state_objects import State
from glue.core import BaseCartesianData, Subset, ComponentID
from glue.utils.matplotlib import MATPLOTLIB_GE_36
from echo import CallbackList, CallbackDict
from matplotlib.colors import Colormap
if MATPLOTLIB_GE_36:
    from matplotlib import colormaps
else:
    from matplotlib import cm as colormaps
```
Thanks for addressing this. At this point there is only a requirement (via glue-core) on matplotlib>=3.2, so the old interface needs to be kept available.

glue_jupyter/state_traitlets_helpers.py
```diff
@@ -74,7 +74,7 @@ def update_state_from_dict(state, changes):
                 if changes[name] != MAGIC_IGNORE and getattr(state, name) != changes[name]:
                     if name == 'cmap':
                         if changes[name] != state.cmap.name:
-                            setattr(state, name, get_cmap(changes[name]))
+                            setattr(state, name, matplotlib.colormaps[changes[name]])
```
dhomeier, line 77:
```suggestion
                            setattr(state, name, colormap.get_cmap(changes[name]))
```
Probably better to stick with the function – although the `state.cmap.name` suggests this will always be called in the string form, `get_cmap` supports all three possible input types, although direct lookup might be faster.
For `colormaps[changes[name]]` to work the legacy import above would have to be changed to
```
from matplotlib.cm import cmaps_listed as colormaps
```
(but then would provide no `get_cmap` at all).

dhomeier (changes requested):
There are no tests for that functionality which may be the reason we missed the deprecation phase, but I've tested the options for matplotlib 3.3.4-3.9.0dev.
There is one other use in `utils.py` that should also be updated along with this; I think `scalar_to_png_data` should support `Colormap` and thus use `get_cmap` even more than here. Thanks!

pllim:
Thanks for the review, @dhomeier ! I think I have applied all your comments. Does it look good now?

dhomeier:
Looks all good, and thanks for the isort cleanup, too!

Issue #450, opened 2024-05-17
Issues with default 3D rendering


astrofrog:
This is the default rendering of the L1448 cube:

![image](https://github.com/glue-viz/glue-jupyter/assets/314716/e22ec575-8ab4-4322-9ece-ff3b75e841c4)

The stretch doesn't look good by default, and it looks like because the cube is not, well a cube, it only fills some of the available volume.

Issue #449, opened 2024-05-17
3D scatter viewer broken


astrofrog:
The 3D scatter viewer is no longer showing any points for me, and also shows an additional panel between the visualization and the regular glue viewer and layer options:

![image](https://github.com/glue-viz/glue-jupyter/assets/314716/e4cd3988-795c-4f48-846a-42232b2e3fcb)

@maartenbreddels - before I dig too far into this, are you aware of any changes in ipyvolume recently that might have caused this?

astrofrog:
To hide the side panel we need https://github.com/widgetti/ipyvolume/pull/442

Carifio24:
@maartenbreddels It must be something in how the scatter is wired up to the glue viewer (though I haven't really had time to investigate), as Azmé and I have made ipyvolume scatters that work, but then the same data won't display in glue.

Carifio24:
@astrofrog @maartenbreddels I figured out the basic issue here . It looks like trying to make a plot where one of the x/y/z components has an underlying numpy array whose data type isn't a float won't work. And it seems to fail in a way that breaks updating the scatter widget, as changing the offending component to something that's floats doesn't fix things

As an example see [this gist](https://gist.github.com/Carifio24/dbcec0985eeef5b1ce826d350c3aa8e1). The original W5 data won't show, but if I create a new data object with non-float components removed, it renders fine. I'm not sure whether this is an ipyvolume issue (some sort of bad serialization?) or whether we need to do some `astype`-ing in the layer artist.

PR #456, opened 2024-07-02
Updates for ipyvolume viewer issues

### LINKED ISSUE #451: Missing viewer options in 3D viewers
The following options appear to be missing in the 3D viewers:

* The ability to change the stretch along x/y/z
* The ability to properly set the aspect ratio to native (there is an option but it doesn't work quite as expected)
* The ability to change from perspective to isometric projection
* The ability to control whether to downsample while panning (not crucial though)

Carifio24:
While Azmé and I were investigating the aspect ratio issue mentioned in #451, we noticed that currently, the displayed volume rendering in the ipyvolume viewer seems to be wrong (in particular, it doesn't match what's displayed in the vispy volume viewer). I believe this might also be a partial cause of #450.

I think this is an issue with axes. As an example, I looked at the reduced TAN C14 cube used in the astropy FITS cube tutorial [here](https://learn.astropy.org/tutorials/FITS-cubes.html). If I open this in vispy, I get
Screenshot_2024-07-01_at_1 38 37_AM_optimized

but with the current ipyvolume viewer I get

Screenshot_2024-07-01_at_1 41 10_AM_optimized

which is what led me to believe that the data was being passed into the ipyvolume widget incorrectly. After rearranging the axes as is done in this PR (note that `np.transpose` returns a view if possible, so we shouldn't need to worry about data getting copied), we get
Screenshot_2024-07-01_at_1 39 19_AM_optimized

This also fixes #451 for me, in that using the native aspect ratio now behaves the same as it does in vispy.

It's hard to tell how the changes to the ipyvolume viewer code  affect the 3D scatter viewer since as #449 points out, it isn't working right now. Only thing I can check is that with these changes, the x/y/z labels indicate that it has a right-handed z-up coordinate system, so that's good.

### DIFF
--- a/glue_jupyter/common/state3d.py
+++ b/glue_jupyter/common/state3d.py
@@ -85,7 +85,7 @@ class ViewerState3D(ViewerState):
         # TODO: this could be cached based on the limits, but is not urgent
         aspect = np.array([1, 1, 1], dtype=float)
         if self.native_aspect:
-            aspect[0] = 1.
+            aspect[0] = 1.0
             aspect[1] = (self.z_max - self.z_min) / (self.x_max - self.x_min)
             aspect[2] = (self.y_max - self.y_min) / (self.x_max - self.x_min)
             aspect /= aspect.max()
--- a/glue_jupyter/ipyvolume/common/viewer.py
+++ b/glue_jupyter/ipyvolume/common/viewer.py
@@ -86,12 +86,10 @@ class IpyvolumeBaseView(IPyWidgetView):
             self._session.command_stack.do(cmd)
 
     def limits_to_scales(self, *args):
-        if self.state.x_min is not None and self.state.x_max is not None:
-            self.figure.xlim = self.state.y_min, self.state.y_max
         if self.state.y_min is not None and self.state.y_max is not None:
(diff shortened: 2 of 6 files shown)


maartenbreddels:
> It's hard to tell how the changes to the ipyvolume viewer code affect the 3D scatter viewer since as #449 points out, it isn't working right now.

Indeed, that was my worry as well.

Carifio24:
Turns out that the scatter viewer does work, albeit with some issues with particular glue components - see https://github.com/glue-viz/glue-jupyter/issues/449#issuecomment-2251228656. After making similar changes to the scatter viewer as well, the display is now looking the same for me using either the ipyvolume scatter or the vispy scatter viewers.
3d-scatter-ipv
scatter-3d-vispy

Carifio24:
@maartenbreddels and I had a brief discussion about the scatter issues at today's glue meeting and decided that the best thing to do is to make sure that we pass floats to the ipyvolume scatter. This PR was already manipulating the scatter widget's x/y/z values, so I decided to just add that fix here as well.

It's also now easy to confirm that the axis changes here lead to a right-handed coordinate system (which is what we want).

astrofrog (approved):
Thanks for all the work on this!

PR #457, opened 2024-08-02
Remove debug panel in ipyvolume viewers

Carifio24:
This PR removes the side debug panel in the ipyvolume viewers. Thanks to @maartenbreddels for the tip on how to do this.

This also increases the default size of the figure itself from 400 -> 600px, as I couldn't see the entire bounding box with a 400px width. Happy to increase the size further if desired.


Carifio24:
Just for context, here's what the view looked like at 400px:
Screenshot from 2024-08-01 15-40-44

and at 600px:
Screenshot from 2024-08-01 15-41-06

astrofrog (approved):
Thank you!

PR #459, opened 2024-08-12
Translate Qt sessions to Jupyter sessions

astrofrog:
Work in progress, seems to work for basic plot types but need to work on 3D viewers and WWT. Also need to decide on an API and a way to select which widgets to use (bqplot vs matplotlib, vispy vs ipyvolume)


astrofrog:
Note to self, if glue-jupyter is installed it will break session loading in glue-qt as the plugin will always apply the patches. I need to find a way of conditionally applying it just when loaded from Jupyter.

PR #461, opened 2024-08-13
Fix issue with contour labels not updating after unit change

astrofrog:
@rosteen - hopefully this will fix the issue you were seeing?


rosteen:
> @rosteen - hopefully this will fix the issue you were seeing?

It did! Contour labels  are now updating as expected on unit conversion. Thanks!

https://github.com/user-attachments/assets/2c08704a-53fd-40cc-8392-95a14156753a

Issue #353, opened 2023-04-18
Scatter empty at initialization


martinRenou:
Originally opened in https://github.com/QuantStack/glue-lab/issues/24

The scatter view is empty by default:

```python
import glue_jupyter as gj
app = gj.jglue()
data_image = app.load_data('w5.fits')

scatter = app.scatter2d(data=data_image)
```

![Screenshot from 2023-04-18 17-10-00](https://user-images.githubusercontent.com/21197331/232821323-a2f6777d-423b-467d-b0d7-7cb8c0cf637d.png)

The scatter view is only populated after some mouse interactions with it (create a subset then delete the subset).

Is this the intended behavior?

martinRenou:
What I thought was a bqplot issue was actually the glue-jupyter implementation https://github.com/glue-viz/glue-jupyter/blob/main/glue_jupyter/bqplot/scatter/layer_artist.py#L49

jfoster17:
> What I thought was a bqplot issue was actually the glue-jupyter implementation https://github.com/glue-viz/glue-jupyter/blob/main/glue_jupyter/bqplot/scatter/layer_artist.py#L49

I don't think that line is the culprit. The general pattern for glue viewers is to create them empty and then redraw them when we add data, and this line just follows that pattern.

Some more data for chasing down this issue; this seems to mainly (only?) affect the first scatter viewer in a notebook. Thus in this video the second scatter viewer shows up properly right away:

https://user-images.githubusercontent.com/3639698/236013470-eeca8a13-68f3-4325-ac2a-91c1bc942938.mov

This makes me suspect that the problem is not with the glue viewer logic -- at least not entirely. Similarly, this problem does NOT occur for me if I simply restart the kernel and re-execute the notebook. 

https://user-images.githubusercontent.com/3639698/236013953-701ab148-2410-479d-80da-838b4becafda.mov

A [potentially related issue](https://github.com/glue-viz/glue-jupyter/issues/243) suggests the problem might also be OS/browser related...

astrofrog:
I've definitely seen something like this in the past - that the first time one shows a plot it is empty but it works when executing the cell again.

Note 100% the same but in case it is related: https://github.com/bqplot/bqplot/issues/856

martinRenou:
Thanks for commenting!

I've explored the bqplot code, and bqplot seems to not receive any data on the front-end when first executed. So it may not be a bqplot issue?

[...]

PR #467, opened 2024-08-26
Disable scatter viewer density map image broadcast when not visible

kecnry:
This PR prevents glue from broadcasting a large binary image for scatter viewers when the density map is not in use.

This still needs testing to ensure that density maps still continue to function as designed.


dhomeier:
I have tested this with the examples shown in #353 and found similar problems to the original issue, namely the density map not reappearing automatically when the `show points` switch is turned off and then back on again:
density_map-toggled
density_map-modified

`density_mark.visible` is immediately set to `True` when `show points` is, but `density_mark.image` appears to remain set to `EMPTY_IMAGE` until some other update is triggering `_update_rendered_image`, like changing contrast, or sometimes accessing some other property of the layer (the latter not very reproducibly).
But we probably need an extra `_update_rendered_image` on changes to `points_mode`.

astrofrog (approved):

PR #472, opened 2024-10-29
Try and improve responsiveness when debounced isn't really needed

astrofrog:
The main motivation for ``@debounced`` is to avoid many repeated updates when e.g. panning/zooming or doing other front-end changes. However, currently, there are cases where relying on debounced is not desirable, e.g.:

```
image.state.reset_limits()
time.sleep(3)
```

In this case, if this code is in a cell, the figure won't be updated until after the sleep because the callback for the debounced update is not in a separate thread. The long-term fix would be to try and make sure glue and glue-jupyter are thread-safe and use a thread in this case, but for now, a simple fix is to simply treat updates in limits similarly to e.g. updates in ``percentile``, ``cmap`` and so on, and make sure those changes result in an immediate update in the figure.

This PR also adds a check to avoid unnecessary repeated updates in ``.update()``


bmorris3:
Thanks @astrofrog, this is helpful 🙌🏻

PR #475, opened 2024-12-11
fix: use last focused/resized view to determine shape of BqplotImageView

iisakkirotko:
# Pull Request Template

## Description

Instead of always using the lowest cid view to determine the aspect ratio / shape of a Bqplot image viewer, use the last focused / resized one. `focused_at` describes the time that the window of that view was last focused, while `resized_at` describes when each viewer was last resized, so this way distortion from resizing based on a view in a popout window will be fixed once the original window is brought into focus.

Needs https://github.com/glue-viz/bqplot-image-gl/pull/112 to work. Let me know if you need this to use the old behaviour as a fallback with previous versions of bqplot-image-gl.

TODO:
- [x] Bump `bqplot-image-gl` in the dependencies once https://github.com/glue-viz/bqplot-image-gl/pull/112 is released.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -36,7 +36,7 @@ test =
     nbconvert>=6.4.5
 visualtest =
     playwright==1.41.2
-    pytest-playwright
+    pytest-playwright==0.5.2
     pytest-mpl
     solara[pytest]
 docs =
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,7 +24,7 @@ install_requires =
     ipywidgets>=7.4.0
     ipyvue>=1.2.0,<2
     ipyvuetify>=1.2.0,<2
-    bqplot-image-gl>=1.5.0
+    bqplot-image-gl>=1.6.0
     bqplot>=0.12.17
     bqplot-gl
     scikit-image
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,7 +38,7 @@ visualtest =
     playwright
     pytest-playwright
     pytest-mpl
-    solara[pytest]<1.29
+    solara[pytest]
 docs =
     sphinx
     sphinx-automodapi
--- a/setup.cfg
+++ b/setup.cfg
@@ -35,7 +35,7 @@ test =
     pytest-cov
     nbconvert>=6.4.5
 visualtest =
-    playwright
+    playwright==1.41.2
     pytest-playwright
     pytest-mpl
     solara[pytest]
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,7 +38,7 @@ visualtest =
     playwright==1.41.2
     pytest-playwright==0.5.2
     pytest-mpl
-    solara[pytest]
+    solara[pytest]>=1.44.1
 docs =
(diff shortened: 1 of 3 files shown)


maartenbreddels:
> Let me know if you need this to use the old behaviour as a fallback with previous versions of bqplot-image-gl.

I think it's fine like this, if we also bump https://github.com/glue-viz/glue-jupyter/blob/1876875940aa876071776b51cf6b017a67702a88/setup.cfg#L26

Otherwise some people might get the old behaviour.

maartenbreddels:
We can use version 1.6.0 now (https://pypi.org/project/bqplot-image-gl/1.6.0/)

dhomeier:
Could you rebase to try and get the tests running?

iisakkirotko:
@dhomeier done!

dhomeier:
@maartenbreddels: visual failure is related to astropy/astropy#17683, right?

maartenbreddels:
Indeed

dhomeier:
Hmm, updating to solara 1.44 seems to have broken more than it fixed – any ideas what those pytest-playwright failures are?

dhomeier:
The [last run prior to upgrading solar](https://app.circleci.com/pipelines/github/glue-viz/glue-jupyter/305/workflows/10c36962-ddfe-4581-acd1-17044f05f71d/jobs/340) already was at `pytest-playwright==0.6.2` so thought it was already pinned somehow. But I think that already raised on an earlier point in collection, so perhaps does not tell us much.

maartenbreddels:
Downgrading pytest-playwright helped, but now I hit https://github.com/widgetti/solara/issues/606 - which I can not reproduce (so working on that now).

setup.cfg
```diff
@@ -35,10 +35,11 @@ test =
...
     nbconvert>=6.4.5
 visualtest =
-    playwright
-    pytest-playwright
+    playwright==1.41.2
+    pytest-playwright==0.5.2
     pytest-mpl
-    solara[pytest]<1.29
+    pytest-ipywidgets[all]
+    solara-ui @ https://github.com/widgetti/solara/archive/refs/heads/fix_matplotlib_backend_keyerror.zip
```
dhomeier, line 42: Something therein is pinning jupyter-client 7.x, jupyterlab 3.x and notebook 6.x now, among others.
But for visualtest only we probably can/have to live with that for now.

dhomeier (approved):
Approving as is, but if you want to pull the CI fixes out to further investigate the frame colour issue @maartenbreddels, fine with me, too.

maartenbreddels:
I've digged into why this is happening, and I think this is not a related change, and we should be able to merge this.

It seems that the output we see now is correct, since I see that bqplot-image-gl gets fed the border with values `0.2 0.2 0.2 1.` which is an opaque dark grey. However, I don't understand why the current reference image is white. This might be a change in glue (maybe @astrofrog can confirm).

maartenbreddels:
@dhomeier if you agree, please rebase and merge, not squash.

PR #477, opened 2025-03-05
WIP: Initial addition of components for point select

CyclingNinja:


glue_jupyter/bqplot/common/tools.py
```diff
@@ -6,6 +6,7 @@
 from bqplot.interacts import BrushSelector, BrushIntervalSelector
 from bqplot_image_gl.interacts import BrushEllipseSelector
 from glue import __version__ as glue_version
+from glue.core.edit_subset_mode import ReplaceMode
```
astrofrog, line 9: Is this needed?

astrofrog (commented):
This seems like it's on the right track for bqplot, but I think maybe it would be easiest to focus on just the bqplot changes here and not do the ipyvolume for now, as that's going to be more challenging to do. If the matplotlib one works, it might also be easier to split out into its own PR?

PR #483, opened 2025-05-27
feat: support for rotated ellipse and rectangle selections

mariobuikhuizen:
## Description

This PR adds support for rotated selections introduced in https://github.com/glue-viz/bqplot-image-gl/pull/127

### DIFF
--- a/.github/workflows/ci_workflows.yml
+++ b/.github/workflows/ci_workflows.yml
@@ -10,7 +10,6 @@ on:
 
 jobs:
   initial_checks:
-    if: false
     # Mandatory checks before CI tests
     uses: OpenAstronomy/github-actions-workflows/.github/workflows/tox.yml@v1
     with:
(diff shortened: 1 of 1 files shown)


dhomeier:
Thanks for straightening the pre-commit out; there are probably some new complaints coming up with the update to ruff 0.12, but might deal with that perhaps in the auto-update PR.

> There is still an issue in the tests: changing the angle of the selection isn’t synced back to the bqplot selection. I can’t figure out how to do that in glue-jupyter. In jdaviz this happens eventually by deselecting and then reselecting a subset. Do you have any idea?

Do you mean how to set `roi.theta` from `BrushRectangleSelector`?

mariobuikhuizen:
> The rotated rectangle reference looks not rotated at all now, or is that the `rotate(180,...)` version?

Good catch, that was an unintended change, I will fix that.

astrofrog:
Minor: can you rename ``selection_test.py`` to ``test_selection.py``? (we try and always use that convention when we can - some older files might not follow this but we'll need to change them).

It would be good ideally to resolve the mismatch in rotation in the sense that we do want the selection ellipse to match exactly what points are selected, but also happy for this to be merged as is even if not perfect and to address that in a follow-up PR.

PR #487, opened 2025-07-09
Add Point Selector tool from glue 1.23 to BQplot image viewer

CyclingNinja:
Adds the matplot lib and BQplot functionality for point selection
 
Includes initial steps for extraction of pixel selection mode to glue-core


CyclingNinja:
!autofix

.github/workflows/ci_workflows.yml
```diff
@@ -50,13 +50,13 @@ jobs:
...
-        - linux: py310-notebooks
+        - windows: py310-notebooks
+        - macos: py311-notebooks
+        - linux: py313-notebooks
 
         - linux: py311-docs
-        - macos: py39-docs
-        - windows: py38-docs
+        - macos: py310-docs
+        - windows: py10-docs
```
dhomeier, line 59:
```suggestion
        - windows: py313-docs
```
I assume this was meant to be `py310`, but could as well try a newer version here.

dhomeier (commented):
Sorry for not testing this more thoroughly downstream, I’ve confirmed non-functioning cases directly in a Jupyter notebook as well!
> Additionally, I'm wondering why brush selection was used here given that the intent was to select a single point - brush selection allows the selection of a region of multiple points, when all that's needed here is registering the location of a single click.

The motivation was probably that there is no frontend method for that unless I’ve been missing something in bqplot-image-gl. But indeed this is less than ideal, as from brushing over a region at best you could set the centre as selected point.

rosteen:
> Apologies rosteen there were a lot of moving parts at the time and this must've slipped through. I'll address this presently

All good, thanks for taking another look!

dhomeier:
@CyclingNinja any updates on this?

PR #488, opened 2025-07-29
Adds dev builds to ci deployment

CyclingNinja:


tox.ini
```diff
@@ -23,12 +22,16 @@ changedir =
 deps =
     notebooks: astroquery
     notebooks: pyyaml
-    devdeps: git+https://github.com/glue-viz/glue
+    dev: git+https://github.com/glue-viz/glue
 extras =
     test: test
+    all: all
```
dhomeier, line 28: What's the `all` group supposed to comprise that you're adding here? I don't see any differences in packages installed or tests run between `py311-test` and `py311-test-all`.

.github/workflows/ci_workflows.yml
```diff
@@ -28,19 +28,26 @@ jobs:
...
-        - windows: py38-test
-        - windows: py39-test
         - windows: py310-test
+        - windows: py311-test
+        - windows: py312-test
+        - windows: py313-test
+
+        # test against some dev versions of some packages
+        - linux: py311-test-all-dev
+        - linux: py311-test-all
```
dhomeier, line 48:
```suggestion
        - linux: py311-test-dev
```
There is no `all` installation target with any different packages.

PR #492, opened 2025-09-29
Replace tornado IOLoop with asyncio

Carifio24:
This PR replaces `tornado` with `asyncio`. The `tornado.IOLoop` class is now a wrapper around an `asyncio` event loop, and we only have the two calls changed here, so I think this should be fine? This is needed for use cases like https://github.com/Carifio24/glue-lite-demo where `tornado` isn't available.

### DIFF
--- a/glue_jupyter/utils.py
+++ b/glue_jupyter/utils.py
@@ -1,7 +1,6 @@
 import os
 import functools
 import collections
-import time
 from io import BytesIO as StringIO
 
 import IPython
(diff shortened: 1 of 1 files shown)


dhomeier:
Thanks, this seems sensible also in light of the deprecation notices on `IOLoop.instance` and `IOLoop.current`.
However we haven't even set an explicit dependency on tornado, so will probably need some more investigation to get an installation without it.
Also needs a rebase as imports have been moved to top level wherever possible in 70054aa.

dhomeier (approved):
I could not follow the exact translation of the `add_timeout` and `add_callback` calls, but everything works, looks good to go.
Amazing work on GluepyterLite!

PR #493, opened 2025-09-30
Debug changing zorder in viewers with mixed layer states

rosteen:
I ran into a bug where changing zorder of the layers in a viewer with both image and scatter layers would error, because `BqplotImageLayerArtist` object has no attribute `density_mark`. This fixes the bug by only accessing each mark in `_update_zorder` if it actually exists on the layer.


rosteen:
It appears that the visual test is failing on other PRs as well and isn't due to this.

PR #494, opened 2025-10-02
Fix subset deletion after reordering layers

rosteen:
This fixes an error I'm hitting in CI after implementing drag-and-drop layer reordering in Jdaviz. It appears that this may be triggered after the image_artist has already been removed due to the reordering, so I added a check that the image_artist is still in the marks before removing it.


dhomeier (approved):
Looks straightforward enough; unless @astrofrog has a better idea to avoid this already in the reordering, good to go, thanks!

PR #495, opened 2025-10-02
Update figure hashes

astrofrog:
The figure for test_contour_units looks better now, also there were hashes missing for two tests.


dhomeier:
Are those the two tests reenabled in #479?

dhomeier (commented):
A bit confusingly some of the tests are passing in #479 with different hashes; here all are [x]passing, but `test_visual_linestyle` and `test_visual_vector` are still marked xfail.

PR #504, opened 2026-02-04
Add dropdown to select which columns are shown

astrofrog:
@kecnry - it was already possible to control this with the viewer state attribute ``state.hidden_components`` but this adds a UI dropdown for it:

Screenshot 2026-02-04 at 11 24 06

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ dist
 
 # Other generated files
 glue/_githash.py
+glue_jupyter/_version.py
 
 # Other
 .pylintrc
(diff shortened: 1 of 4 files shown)


astrofrog:
@rosteen - well it appears in the viewer options here, so I'm not sure, it depends if you are using a custom viewer options widget? I haven't tested this out inside jdaviz.

rosteen:
Hm yeah I don't think we currently expose viewer options like this at all. @kecnry any idea off the top of your head if we'll need to tie this into Plot Options or show it in the viewer/viewer toolbar somehow?

kecnry:
I think we could just trigger the same `state.hidden_components` with our own widget in plot options then (since otherwise this will be outside the widget we pull in, I think).  Would it be possible to put into the table header, which I think is a pretty common design-pattern?

astrofrog:
Sounds good, I'll just merge this since it might be useful to some, and if you find a good UI in jdaviz let me know and maybe we can port it back here

PR #507, opened 2026-02-05
Removed unused dependencies

astrofrog:
I think none of these are needed anymore


dhomeier (approved):
They are pulled in through some of the other dependencies anyway, so probably makes no difference.

PR #515, opened 2026-02-14
Provide a way to edit cells in table viewer

astrofrog:
To use this, one needs to set ``editable_components`` on the viewer state:

```
table.state.editable_components = [points.id['x']]
```

and when hovering over values a little pen will appear - clicking on it will allow the cell to be edited (see x column):

Screenshot 2026-02-14 at 22 54 03

Changes are currently automatically synced to the ``Data`` object, which is not great performance-wise so we should think whether we want to have a way to only sync on request

Open questions:

* Do we want editing a cell to actually bring up a modal dialog? This might be better if the content to type is long? The current UI has some advantages though.
* Do we want edits to be synced straight away to the rest of glue? This might be slow for datasets with millions of rows, so it just depends what kind of datasets we will use this with

### DIFF
--- a/glue_jupyter/table/tests/test_table.py
+++ b/glue_jupyter/table/tests/test_table.py
@@ -258,3 +258,97 @@ def test_table_apply_subset_tool(app, dataxyz):
     subset = table.layers[1].layer
     mask = subset.to_mask()
     assert list(mask) == [True, False, True]
+
+
+def test_table_editable_components(app, dataxyz):
+    """Test that setting editable_components marks headers as editable."""
+    table = app.table(data=dataxyz)
+
+    # Initially no columns are editable
+    assert table.state.editable_components == []
+    headers = table.widget_table.headers
+    for header in headers:
+        assert header['editable'] is False
+
+    # Set editable components using ComponentID
+    table.state.editable_components = [dataxyz.id['x']]
+
+    # Check header has editable flag
+    headers = table.widget_table.headers
+    x_header = next(h for h in headers if h['value'] == 'x')
+    assert x_header['editable'] is True
+
+    # Other columns should not be editable
+    y_header = next(h for h in headers if h['value'] == 'y')
+    assert y_header['editable'] is False
+
+    # Test multiple editable components
+    table.state.editable_components = [dataxyz.id['x'], dataxyz.id['y']]
+    headers = table.widget_table.headers
+    x_header = next(h for h in headers if h['value'] == 'x')
+    y_header = next(h for h in headers if h['value'] == 'y')
+    z_header = next(h for h in headers if h['value'] == 'z')
+    assert x_header['editable'] is True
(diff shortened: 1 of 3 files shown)


dhomeier:
> * Do we want editing a cell to actually bring up a modal dialog? This might be better if the content to type is long? The current UI has some advantages though.

What do you have in mind for this? The current one is straightforward, though a dialog might provide some info on the valid content – currently entering incompatible input just fails silently. Should usually be clear enough to infer from the existing values.

One problem I ran into after you added the cross and tick though is that in very narrow columns I can no longer see what I'm typing at all – the ✔️is already covering the full field.

dhomeier:
Seems to work well enough when the cell is wide enough. But otherwise typing blindly is error-prone, since by default input seems to be inserted over the first position only (e.g. I type `IV` to replace a `III` and get `IVII`).

Also, how exactly is the input parsed? I was playing with the `np.float(nan)` fields in the W5 table, which accepts numerical input as it should, but I am unable to enter something like `np.nan` or `np.inf` again.

notebooks/Generic/demo_gaussian.ipynb
```diff
@@ -2,17 +2,33 @@
...
-   "metadata": {},
+   "execution_count": 1,
+   "metadata": {
+    "execution": {
+     "iopub.execute_input": "2026-02-17T21:35:26.261864Z",
+     "iopub.status.busy": "2026-02-17T21:35:26.261407Z",
+     "iopub.status.idle": "2026-02-17T21:35:27.563040Z",
+     "shell.execute_reply": "2026-02-17T21:35:27.562593Z",
+     "shell.execute_reply.started": "2026-02-17T21:35:26.261827Z"
+    }
```
dhomeier: Should this be cleaned up for the version in repo?

glue_jupyter/table/tests/test_table.py
```diff
@@ -198,3 +198,97 @@ def test_table_add_remove_data(app, dataxyz, dataxz, data_empty):
...
+    table = app.table(data=dataxyz)
+    table.state.editable_components = [dataxyz.id['x']]
+
+    original_value = dataxyz['x'][0]
+
+    # Try to edit with invalid value
+    table.widget_table.vue_cell_edited({
+        'row': 0,
+        'column': 'x',
+        'value': 'not_a_number'
```
dhomeier: Might want to also test that `'value': 'nan'` actually works (for a float column)

astrofrog: Good idea

dhomeier (approved):
Works fine for me. From a user perspective I could just think of two more features that might be helpful:
1. an interactive option to make columns editable, like from the context menu (though I have no idea at what level that is implemented)
2. a keyboard shortcut, maybe shift-up/down for moving to next or previous row within the edit cell.

PR #516, opened 2026-02-15
Improve UI for selecting rows

### LINKED ISSUE #429: Improve how selections are made in Table viewer
At the moment, the method to make selections in the table viewer is unintuitive - selecting the checkboxes is fine, but then knowing to click on the filter icon is not. We should try and improve the UI for making selections. At the same time it would be nice to make it possible to select multiple rows in one go - in Qt we can click on a row then shift click on another one lower down to select a range.

astrofrog:
This is to fix https://github.com/glue-viz/glue-jupyter/issues/429 

Still needs tests.

Screenshot 2026-02-15 at 21 39 54

### DIFF
--- a/glue_jupyter/icons/row_select.svg
+++ b/glue_jupyter/icons/row_select.svg
@@ -1,0 +1,121 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   width="100mm"
+   height="100mm"
+   viewBox="0 0 100 100"
+   version="1.1"
+   id="svg5"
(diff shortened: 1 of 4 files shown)


kecnry (approved):

dhomeier:
It's probably impossible to predict what interface works best for everybody's workflow, but when I saw the master selector icon changed to `–` after selecting some rows, I expected it to **de**select all rows again. Just takes one extra click this way to select all, then unselect, but that may be the more common operation. My suggestion for the different states would be
- no rows selected: icon ➕, action `select all`
- some rows selected: icon ➖, action `unselect all`
and as an optional bonus
- all rows selected: icon ✔️, action `unselect all`

@kecnry what do you think is most intuitive?

And +1 for providing some option to select/unselect a range of rows, but that could be a follow-up enhancement.

PR #519, opened 2026-02-24
feat: ipyvuetify 3 compatibility

mariobuikhuizen:
# Pull Request Template

## Description

Adds ipyvuetify 3 compatibility


glue_jupyter/table/table.vue
```diff
@@ -79,6 +80,37 @@ module.exports = {
...
+      const pageSize = Number(value);
+      if (!Number.isFinite(pageSize) || this.items_per_page === pageSize) {
+        return;
+      }
+      this.items_per_page = pageSize;
+      this.options = {
+        ...this.options,
+        itemsPerPage: pageSize,
+        totalItems: this.total_length,
+      };
```
astrofrog: Just curious, is this adding a new feature or is it that what we had before is no longer possible by default?

astrofrog (commented):
Thanks! Just a comment below. I think we should probably also hold back from merging this until ipyvue 3.0.0 final is released?

glue_jupyter/table/tests/test_table.py
```diff
@@ -93,7 +93,7 @@ def test_table_column_visibility_widget(app, dataxyz):
     options_widget = table._layout_viewer_options
 
     # Check that all columns are listed and visible by default
-    column_names = [item['text'] for item in options_widget.column_items]
+    column_names = [item['title'] for item in options_widget.column_items]
```
dhomeier, line 96: Probably requires the expected values in `test_vuetify_helpers.py` to be changed accordingly to `{'title': 'A', 'value': 0}` etc. to match.

dhomeier:
Ah, the visual failures are indeed related to the update – looks like the viewer options are now opened by default:
result

dhomeier (changes requested):
I have the same problem with the viewer options already opened in the example notebook, and moreover can't seem to find a way to close it at all.

mariobuikhuizen:
I'm looking into it...

mariobuikhuizen:
@dhomeier The only visual difference in the ui tests is that the toolbar looks different, because the look of toggle buttons has changed in Veutify 3. Do you think this visual change is acceptable?

Screenshot 2026-09-01 at 13 58 38

dhomeier:
I doubt many people would notice the difference on update. Don't know if it would be possible to set the default button background back to white again, but if those are design choices, should be perfectly acceptable on a major backend update. That would probably just require updating the [hashes](https://app.circleci.com/pipelines/github/glue-viz/glue-jupyter/789/workflows/5c24fc03-6b59-4af1-9fe0-79e9315d3bca/jobs/1054#:~:text=251-,252) for the _selection tests.

astrofrog:
@mariobuikhuizen besides the hashes which I can update, is this ready for a final review/merging?

dhomeier:
The idea was to get glue-viz/echo#72 merged and released and then update the echo minver here, right?

astrofrog:
Sounds good! Sorry I've been out of the loop

pyproject.toml
```diff
@@ -25,15 +25,14 @@ authors = [
 ]
 
 dependencies = [
-    "echo>=0.14.0",
+    "echo @ git+https://github.com/glue-viz/echo.git@refs/pull/72/head",
```
dhomeier:
```suggestion
    "echo>=0.16.0",
```
released and on PyPI now

dhomeier (approved):
Tests are passing with the new echo release.

astrofrog:
Ok but we need to update the visual hashes, I'll do this now

astrofrog:
Oh I see all the failing ones have no hashes so we just need to merge to regenerate the baselines

PR #520, opened 2026-02-24
Update ipyvolume viewer to use reference data

Carifio24:
This PR updates the ipyvolume viewer to use the same viewer-level resolution/fixed resolution buffer scheme used in e.g. `glue-vispy-viewers`.

The one thing that I'm having trouble with here is the float sliders that control vmin/vmax for each layer. It looks like they should be set to track the min/max data from the layer state, but they keep getting reset to 0/1.


astrofrog:
@Carifio24 - just to check, is this ready for review/testing?

astrofrog (changes requested):

Carifio24:
@astrofrog I was able to work out the warning issue I was having with the scatter viewer widget (all it took was using the `skip` kwarg for the autoconnections).

glue_jupyter/ipyvolume/volume/layer_style_widget.vue
```diff
@@ -0,0 +1,57 @@
...
+    <div>
+      <v-slider label="Opacity" v-model="alpha" :min="0" :max="1" :step="0.001" hide-details />
+    </div>
+    <div>
+      <v-slider label="Opacity scale" v-model="opacity_scale" :min="-3" :max="3" :step="0.01" hide-details />
+    </div>
+    <div>
+      <v-select label="Stretch" :items="stretch_items" v-model="stretch_selected" hide-details />
+    </div>
+    </div>
```
astrofrog, line 44: Duplicate </div>

glue_jupyter/ipyvolume/volume/layer_style_widget.vue
```diff
@@ -0,0 +1,57 @@
...
+    <div>
+      <v-subheader class="pl-0">lighting</v-subheader>
+      <v-switch v-model="lighting" hide-details style="margin-top: 0" />
+    </div>
+    <div>
+      <v-select label="Render method" :items="render_method_items" v-model="render_method_selected" hide-details />
+    </div>
+    <div>
+      <glue-float-field label="vmin" :value.sync="vmin" echo-type="float" />
+      <glue-float-field label="vmax" :value.sync="vmax" echo-type="float" />
```
astrofrog: This seems to duplicate the vmin/vmax values on lines 29-32?

glue_jupyter/ipyvolume/volume/layer_artist.py
```diff
@@ -58,11 +72,14 @@ def __init__(self, view, viewer_state, layer=None, layer_state=None):
                                  tf=self.transfer_function, data_max_shape=128)
         self.figure.volumes = self.figure.volumes + [self.volume]
 
+        self.id = str(uuid.uuid4())
```
astrofrog, line 75: Is this used anywhere? If so I think ``self.uuid`` would be better to not shadow the built-in ``id``

glue_jupyter/ipyvolume/scatter/layer_style_widget.vue
```diff
@@ -0,0 +1,70 @@
...
+      <template v-if="vector_visible">
+        <v-select label="vx" :items="vx_att_items" v-model="vx_att_selected" />
+        <v-select label="vy" :items="vy_att_items" v-model="vy_att_selected" />
+        <v-select label="vz" :items="vz_att_items" v-model="vz_att_selected" />
+      </template>
+    </div>
+  </div>
+</template>
+
+<style id="viewer_3d">
```
astrofrog: I think this ``id`` is also used in ``viewer_options_widget.vue``, worth using a different id, or using a class?

astrofrog (changes requested):
Just a few small comments, but looks good to go otherwise!

glue_jupyter/ipyvolume/volume/layer_style_widget.vue
Carifio24: Sorry but where is there another stretch widget? I'm not seeing it

glue_jupyter/ipyvolume/volume/layer_style_widget.vue
Carifio24: Ah yeah, these are duplicates. I also updated so that the remaining widgets don't appear for a subset.

glue_jupyter/ipyvolume/scatter/layer_style_widget.vue
Carifio24: You're right, this is a duplicate. I changed to have a distinct ID. I also gave both layer style widgets a common `layer_3d` class in case that's useful.

glue_jupyter/ipyvolume/volume/layer_artist.py
Carifio24: It's actually used in `glue-core` for the creation of the FRB: https://github.com/glue-viz/glue/blob/main/glue/viewers/volume3d/data_proxy.py#L110, so that it can be checked against what's in the array and pixel caches

PR #527, opened 2026-03-17
Enable keyboard navigation from the table edit cell

dhomeier:
## Description
#525 – this is adding support for the up/down keys to move forward or back a row.
I did not restrict the movement to editable cells, so the actions will be
1. `Escape`: leave cell
2. `Enter`: if `editable`, commit changes, otherwise just move to next row
3. `Up`/`Down`: move to previous/next row

My js coding is extremely limited, I just used Up and Down without modifiers here – this will intercept the navigation between notebook cells, but only when the edit cell is in focus – is this acceptable?
Will return from the cell upon reaching bottom or top of a page, but that is consistent with the behaviour from #515.

In principle this might even be extended to navigating between columns, but the left/right keys are already in demand for in-cell movement. With modifiers like alt-left, this could become usable.

Also added some more edit tests for different inputs.

### DIFF
--- a/glue_jupyter/conftest.py
+++ b/glue_jupyter/conftest.py
@@ -11,6 +11,11 @@ def dataxyz():
 
 
 @pytest.fixture
+def datamix():
+    return Data(x=[1, 2, 3], y=[2.3, 3.14, 4], z=['a', 'b', 'c'], label="mixed type data")
+
+
+@pytest.fixture
 def datacat():
     return Data(a=['a', 'b', 'c'], b=['d', 'e', 'f'], label="categorical data")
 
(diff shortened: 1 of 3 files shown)


dhomeier:
Marking this as ready as I haven't heard any suggestions for different functionality.

dhomeier:
@astrofrog do you think this is ready to go in?

astrofrog (approved):

dhomeier:
Sorry for the belated commit @astrofrog ; in final testing against current head I noticed that for non-editable cells the focus does not switch automatically to the edit field on clicking the cell, so using the keyboard navigation requires an extra click into the field on top. That commit changes this to consistent behaviour across editable and non-editable cells. This _could_ interfere with the keyboard shortcuts for cell manipulation (like `C`: Copy, `V`: Paste); however I have not been able to use those even with the main branch and am not sure how they are supposed to work. But if you think it's problematic, I can just revert the last commit.

PR #528, opened 2026-03-19
Added support for log axes in scatter viewer

astrofrog:
This is an alternative to https://github.com/glue-viz/glue-jupyter/pull/273


astrofrog:
The log labels are sometimes a bit ugly but for now I'm just going with bqplot defaults and will open a follow-up PR to try and improve things.

dhomeier:
Testing this with a regular notebook I find that the content is no longer updated (i.e. stays in either linear or logarithmic scale) after switching back and forth between the options a couple of times. Will also test in jupyterlab.

astrofrog:
@dhomeier - can you check if the issue you saw has gone away? If not, can you send a screen recording?

dhomeier:
Yes, updating works in both axes as many times as you switch, as far as I can tell.
One minor remaining issue is that on drawing a subset, the gray shaded area remains always drawn in linear coordinates (but the points in subset, as shown by colour, are correctly displayed).
Similar to problems already seen whenever axes have different types, and probably difficult to resolve as e.g. an elliptical ROI is very non ellipse-like in semilog coordinates.

Screenshot 2026-04-22 at 12 25 22 am

Another issue I just noted is that pan-and-zoom becomes blocked for any log axis (or at least seems so, again maybe not too surprising) and remains in that state even when switching that axis back to linear.

dhomeier:
zooming and panning remains usable with the latest commits; original viewport is now reset on any switch of an axis' mode, but can be modified in log scale as well. I guess resetting is unavoidable or even preferable when changing the scale, but doing this on a per-axis case has a confusing side-effect when changing only one axis:
the axis that is switched from lin to log or back has its viewport reset, while the other keeps its modified range.
This can actually be used to zoom in one axis only (haven't found a modifier that would do this otherwise), but probably still comes around somewhat unexpected.

astrofrog:
Yes sorry this still in the middle of some changes, I am now trying to resolve a bqplot bug

astrofrog:
pre-commit.ci autofix

Carifio24:
@astrofrog I'm running into an issue (which it seems like had been seen earlier?) with the scatter mark seemingly not updating properly when I change the log flag for the axis.

As an example, I set up a viewer with
```python
N = 10
x = list(range(N))
y = [10 ** t for t in x]
data = Data(label="Test Data", x=x, y=y)
app.data_collection.append(data)

sv = app.new_data_viewer(BqplotScatterView, data=data)
layer = sv.layers[0]
layer.state.size = 25
layer.state.size_scaling = 4
layer.state.color = "#0000ff"
```

When I change the "log y axis" flag, the marks on the axis update how I'd expect but the scatter mark itself doesn't update accordingly (should be a straight line with a log y axis). Similar for the log x axis, the points look uniformly spaced. LMK if I'm setting this up wrong somehow.


https://github.com/user-attachments/assets/411a514c-3c0f-406c-9c4a-9cb99cf7f376

astrofrog:
@Carifio24 - strange, that example works for me. What versions of bqplot, bqplot-image-gl and bqplot-linlog do you have installed?

astrofrog:
I wonder if you have an old version of something, because for me I see minor ticks on the y log scale - could you try making sure all your bqplot packages are up to date?

Carifio24:
@astrofrog I had bqplot 0.12.45, after updating to 0.12.47 this is working as expected!

I tried with 0.12.46 and that worked fine, but with 0.12.45 I see this issue on both my Macbook and Ubuntu machine. Maybe we need to adjust the pin?

astrofrog:
Yes sounds good!

astrofrog:
I think this seems to be working ok and @Carifio24 confirmed it works fine with the most recent bqplot, so going to go ahead and merge this.

dhomeier:
@astrofrog I think I could bisect the bug reported in spacetelescope/jdaviz#914 to d3e8af8, so it seems there is either some incompatibility with bqplot-linlog, or something broke in the code simplification.

astrofrog:
Any console errors?

dhomeier:
None; I suspect it's the calls to `LinLogScale` and `LinLogAxis` already breaking on initialising the window.
Interestingly I can get a functional scatter plot window with the _Display in new window_ button.
@MatthewPortman can you confirm if the pop-out window works for you as well?

dhomeier:
Works in jupyterlab, too.

astrofrog:
@dhomeier - when you say works in Jupyter Lab, you mean the issue happens only with vanilla notebooks?

astrofrog:
I'm confused, as everything seems to work for me in notebook and lab, I'm using this notebook: https://gist.github.com/astrofrog/be00e9a3a756aebff6be33de221fd29b - can you point me to how to get it to fail?

dhomeier:
Yes. Or used to, at least – currently I am at a loss to reproduce it at all: I tried some more setups, then tested the same notebook with Firefox (compared to Safari on macOS), but somewhere at that point it no longer occurred in the standard notebook in Safari either. I created a fresh vanilla env with `pip install jdaviz notebook`, still no trace of it.

astrofrog:
There was an incompatibility with bqplot<0.12.46 which caused scatter plots not to show, maybe it was that? (a fresh install would pull in a more recent version)

dhomeier:
I had bqplot 0.12.47 installed for most or all of this, with no evident change in behaviour.
Was using @MatthewPortman's example from Slack, which is using `ldr = jd.loaders['object']` and then `ldr.importer` for loading, but otherwise seems to work the same with your example (i.e. no missing windows anymore). I also did have an issue with the plot markers not showing up in the Safari inline plot, which no longer occurs either. Some stuff just cleared from cache???

MatthewPortman:
I've now tried both my working environment and a clean install but the issue no longer appears. 

For what it's worth, my working environment (where I first saw the issue) has bqplot 0.12.46 while the clean install has 0.12.47.

PR #536, opened 2026-04-21
Respect layer visibility in table viewer

astrofrog:
This makes it so that the visibility of layers is respected in table viewer. Clicking on the eye toggle in the layer list will now show/hide subsets or the overall data, consistent with what we do in glue-qt. If there is a subset, hiding the main dataset will result in the table viewer showing only the subset of rows, which can be useful.

This also fixes a bug that caused the layer color to be incorrect in the layer dropdown list.

cc @kecnry

### DIFF
--- a/glue_jupyter/table/table.vue
+++ b/glue_jupyter/table/table.vue
@@ -243,6 +243,10 @@ module.exports = {
     background-color: #E3F2FD;
 }
 
+.glue-data-table {
+  min-height: 250px;
+}
+
 .glue-data-table .v-data-table__wrapper {
   overflow-x: auto;
 }
--- a/glue_jupyter/table/tests/test_table.py
+++ b/glue_jupyter/table/tests/test_table.py
@@ -352,3 +352,45 @@ def test_table_cell_edit_invalid_value(app, dataxyz):
 
     # Data should remain unchanged
     assert dataxyz['x'][0] == original_value
+
+
+def test_table_visibility_subset(app, dataxyz):
+    table = app.table(data=dataxyz)
+    app.subset('test', dataxyz.id['x'] > 1)
+    assert len(table.layers) == 2
+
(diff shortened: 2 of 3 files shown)


dhomeier:
I am not sure what the customary behaviour is in glue-qt, just noted this detail coming unexpected to me:
when displaying the full table, clicking on the eye lets the rows not in the subset disappear, just as I did expect.
But when I also select the subset in the pull-down menu, the only effect of toggling the eye is removing the subset markers for the rows (screenshots below).

Screenshot 2026-05-05 at 6 08 32 pm
Screenshot 2026-05-05 at 6 10 24 pm

Possibly more confusing, when switching back to full table view from the latter state, the markers remain absent, and now toggling the eye removes all rows:
Screenshot 2026-05-05 at 6 20 03 pm
Screenshot 2026-05-05 at 6 22 21 pm

astrofrog:
Would it be easy to do a screen recording? (No worries if not, just in case you were set up to do it easily)

dhomeier:
Cmd-Shift-5 is your friend ;-) if it helps

https://github.com/user-attachments/assets/829b3b8b-a98e-41e6-806c-9845c5b04939

astrofrog:
> But when I also select the subset in the pull-down menu, the only effect of toggling the eye is removing the subset markers for the rows (screenshots below).

Just to check, what did you expect? It's intentional that if you hide a dataset, the data not in subsets gets removed, and if you unselect just a subset, all rows are still there, because the full data is still visible. However if you had 2 subsets and you hid the data *and* one of the subsets, it should hide the rows that don't appear in the visible subset. Does this all make sense?

astrofrog:
I'll go ahead and merge this but feel free to ping me if you think the behavior is still not quite right or inconsistent with Qt - thanks!

PR #539, opened 2026-05-07
Pin bqplot to <0.13 for now since we are not fully compatible yet

astrofrog:
Things aren't working quite right yet with bqplot 0.13 and bqplot-gl, so pinning this and planning to do a release to avoid people running into issues.


dhomeier (approved):
Would this be a bugfix release? The doc failures would probably need 5842196?

astrofrog:
Yes I think I'll split out that commit into its own PR and once that's fixed we can indeed do a bugfix release.

PR #542, opened 2026-05-17
Update slice helper to support 3d viewers

Carifio24:
This PR extracts out the slice helper-specific pieces of #541.

### DIFF
--- a/glue_jupyter/common/slice_helpers.py
+++ b/glue_jupyter/common/slice_helpers.py
@@ -29,6 +29,7 @@ class MultiSliceWidgetHelper(object):
         self._reference_data = None
         self._x_att = None
         self._y_att = None
+        self._z_att = None
 
         self.sync_sliders_from_state()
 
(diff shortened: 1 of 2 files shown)


glue_jupyter/common/slice_helpers.py
```diff
@@ -1,7 +1,7 @@
 # NOTE: The following MultiSliceHelper is adapted from the Qt version and there
 # is enough overlap that we could consider having a base class for the two.
 
-from ipywidgets import IntSlider
+from ipywidgets import IntSlider, VBox
```
astrofrog: So just to check, are we still planning to use ipywidgets for this?

Carifio24: @astrofrog Good point, we should be using ipyvuetify widgets instead. I've marked this as a draft, will mark as ready for review once I've made that change

Carifio24:
@astrofrog I've updated the slice helper to use the Vuetify sliders. I tried using ipyvuetify elements instead of the `VBox` as the container, but I could not get any of them to display their children, so I just stuck with `VBox`. Since the `VBox` here is just a layout I don't think it breaks the visual consistency.

astrofrog (approved):
Thank you!

PR #548, opened 2026-06-09
[codex] document and test bqplot toolbar compatibility

maartenbreddels:
## Summary

Current `main` already disables bqplot's integrated figure toolbar when `bqplot.Figure` exposes `display_toolbar`. This PR keeps that behavior and makes the compatibility path more explicit:

- check the trait API with `bqplot.Figure.class_traits()` instead of using `hasattr`
- add an inline comment explaining the bqplot 0.13 integrated toolbar change and why glue-jupyter disables it
- add a regression test that simulates a bqplot-0.13-style `Figure` where `display_toolbar` defaults to `True`

The behavior remains the same for users: bqplot 0.12 constructs figures without the unknown keyword, and bqplot 0.13 gets `display_toolbar=False` so glue-jupyter's own toolbar remains the only toolbar shown.

## Tests

- `/tmp/glue-jupyter-bqplot-toolbar-venv/bin/pytest glue_jupyter/bqplot/tests/test_bqplot.py -q` with bqplot 0.13.1: `19 passed, 1 deselected`
- `/tmp/glue-jupyter-bqplot-toolbar-venv012/bin/pytest glue_jupyter/bqplot/tests/test_bqplot.py -q` with bqplot 0.12.47: `19 passed, 1 deselected`

## Note

I attempted to create a signed commit, but this environment has no GPG secret key for `Maarten Breddels <maartenbreddels@gmail.com>`, so the commit in this PR is unsigned.

### DIFF
--- a/glue_jupyter/bqplot/common/viewer.py
+++ b/glue_jupyter/bqplot/common/viewer.py
@@ -41,8 +41,13 @@ class BqplotBaseView(IPyWidgetView):
                              animation_duration=0,
                              axes=[self.axis_x, self.axis_y],
                              fig_margin={'left': 60, 'bottom': 60, 'top': 10, 'right': 10})
-        if hasattr(bqplot.Figure, 'display_toolbar'):
+
+        # bqplot 0.13 added an integrated Figure toolbar that is shown by
+        # default. Glue provides its own toolbar, and bqplot 0.12 does not know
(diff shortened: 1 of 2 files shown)


maartenbreddels:
already added in https://github.com/glue-viz/glue-jupyter/pull/532

PR #129, opened 2019-09-17
feat: add table viewer based in ipyvuetify

maartenbreddels:
Although ipysheet was discussed as a table viewer, I don't think it is the proper widget to do this. It does not enforce a common datatype per column, since it models a spreadsheet, not columnar data. Also, it will be limited in the amount of data it can handle (e.g. don't expect to view the Gaia dataset with it).

Experimenting with vuetify's data table, I've been able to put together a table viewer that will do kernel side pagination, is consistent with the ipyvuetify look, and more easy to customize to our needs.

Note that when we move to vuetify 2.0+, [the data table also has a dense layout](https://vuetifyjs.com/en/components/data-tables#dense)

Example usage (rebased on the vuetify layout branch #115)
table

Example using ~1 billion rows (Gaia data):
image

PR #552, opened 2026-07-15
opt-in column rename/remove from table header UI

kecnry:
This PR implements icons in the table-viewer column headers to rename and remove columns from the data-collection entry for the corresponding table(s).  This functionality is opt-in using the same pattern as `editable_colums`.

https://github.com/user-attachments/assets/ea3102bd-fa4d-4319-b0bf-b7816aefa006

### DIFF
--- a/glue_jupyter/table/table.vue
+++ b/glue_jupyter/table/table.vue
@@ -56,8 +56,7 @@
         :class="['elevation-1', 'glue-data-table', scrollable && 'glue-data-table--scrollable']"
         :style="scrollable && height != null && `height: ${height}`"
       >
-      <template v-slot:header="props">
-        <thead>
+      <template v-slot:headers>
           <tr>
             <th :style="'padding: 0 10px; width: '+Math.max(1, Math.ceil(Math.log10(total_length)))*20+'px'">#</th>
             <th style="padding: 0 1px; width: 30px" v-if="selection_enabled">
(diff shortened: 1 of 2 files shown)


glue_jupyter/table/viewer.py
```diff
@@ -31,6 +33,20 @@ def is_editable(self, component_id):
...
+            if component_id is cid:
+                return True
+        return False
+
+    def is_removable(self, component_id):
+        """Check if a component can be removed from the table."""
+        for cid in self.removable_components:
+            if component_id is cid:
+                return True
+        return False
```
kecnry: this is failing the ruff check but just follows the same pattern as `is_editable` above - what would you prefer?

dhomeier: Apparently `editable_cid in self.editable_components` is just spared because of the line length, so you might thwart ruff by using a longer variable name as well.
But `return any(component_id is cid for cid in self.certain_components)` sure looks more concise to me.
For that matter, is this any different from `return component_id in self.certain_components`, or am I missing something in `ListCallbackProperty` behaviour, @astrofrog?

kecnry: ok, should I change to the more condensed syntax and also update the editable components one to match then?

dhomeier: I've checked that
```python
        return component_id in self.editable_components
```
etc. passes all tests as well, but would wait for @astrofrog to confirm (unless you know of a reason why this cannot be be used instead of the `return any(...)`)

dhomeier:
Unfortunately I am unable to test the actual functionality, since with this PR my table viewer no longer shows column headings (with the W5_Tutorial dataset)! Looks like something must be broken in `table.vue`.

glue_jupyter/table/viewer.py
astrofrog: I think that should be fine!

glue_jupyter/table/table.vue
```diff
@@ -56,7 +56,7 @@
         :class="['elevation-1', 'glue-data-table', scrollable && 'glue-data-table--scrollable']"
         :style="scrollable && height != null && `height: ${height}`"
       >
-      <template v-slot:header="props">
+      <template v-slot:headers>
```
dhomeier, line 58:
```suggestion
      <template v-slot:header="props">
```
  @kecnry with that change reverted the column headers show up again in my example glue-jupyter notebook (plain notebook as well as jupyterlab), and editing and removing is functional as far as I tested. As I said, beyond that I understand very little of this code – I heard in the tag-up it may be related to the latest Jdaviz updates.

glue_jupyter/table/table.vue
dhomeier: @kecnry I noticed that this line is another difference between our upstream version and the copy you initially added in your [Jdaviz PR](https://github.com/spacetelescope/jdaviz/pull/4282/changes/843c3624e5d0cb73f20264c4b0ec7ecfd262870a#diff-a532c65539dc8593c47f6251884bf3033e1a0a5a676555a528459f5f16dfc8f3) – with Jdaviz 5.0.3 and this PR I no longer see the column headings either:
Image
I have not tested against spacetelescope/jdaviz#4282 head, but I find that removing the `hide-default-header`, alternatively to reverting to `v-slot:header="props"` below, also brings the headings back. @astrofrog do you have a suggestion how to proceed here?

kecnry: removing this seems to be the more stable fix (otherwise the header seems to become larger and with a transparent background and would need additional styling to fix 🤷‍♂️ )

dhomeier: Pinging @maartenbreddels as the `hide-default-header` had originally been introduced with #129

PR #553, opened 2026-07-16
fix throttled-slider for vue3 support

kecnry:
This PR fixed glue-throttled-slider to support vue3 (not sure if we need other logic to support _both_ vue2 and vue3 or if this should be part of a larger effort to migrate to vue3 entirely).


dhomeier:
@kecnry since the fix is also included in #519, can this be closed?

